PR #17, opened 2020-11-19
Build with dpNP only on Linux

PokhodenkoSA:
`dpNP` is not available for Windows yet.

Also, `numba-dppy` does not support osx.


PokhodenkoSA:
Build on Windows passed without `dpNP`.

Issue #26, opened 2020-11-02
Add GDB* support to Numba generated GPU kernels.


diptorupd:
Currently, Numba-DPPY has not debugger support for GPU kernels. This feature request tracks the tasks that should be completed to enable debugger support in Numba-DPPY.

- [x] Evaluate what DWARF/debug symbols dpcpp generates for kernels.
- [x] Add flags to Numba-DPPY compilation to generate the same debug symbols for kernels as DPCPP when user has specified "debug=True".
- [x] Work with application debugger team (Matti Puputti) to identify steps on how to run Numba with GDB*
- [x] Develop examples to demonstrate how the debugger should be used.

akharche:
Extended by https://github.com/IntelPython/numba-dppy/issues/149

PokhodenkoSA:
@diptorupd What examples did you mean in checkbox "Develop examples to demonstrate how the debugger should be used"?

PR #27, opened 2020-11-24
Add flags to generate debug symbols

akharche:
Enable DWARF info generating for gpu-kernels: issue https://github.com/IntelPython/numba-dppy/issues/26


PokhodenkoSA (approved):
LGTM. But I would like to see some docs about how developers could use it.
It could be a part of CONTRIBUTION.md file in section: Debugging.
The file is not there yet. You can see on similar one for dpCtl.

PR #44, opened 2020-12-03
Update to new API of dpCtl 0.4.0 related to USM

### LINKED ISSUE #33: Migrate Numba_dppy to dpctl 0.4.0
The memory manager module of `dpctl` was refactored and there are changes to API and paths of the functions. `Numba_dppy` should be updated to uses the latest API of `dpctl.`

@akharche please discuss with @reazulhoque and start with this task.

### LINKED ISSUE #41: Detection of USM allocated arrays should be based on `__sycl_usm_array_interface__`
Related to #33 

https://github.com/IntelPython/numba-dppy/blob/49a172dd908856c0bebc1017b1cf02eafb1004a6/numba_dppy/compiler.py#L425

The following code should be modified to rely on the object `val` to define `__sycl_usm_array_interface__`. An example of such an  object will be the upcoming `dpctl.dptensor.dparray` class. 

The current code handles the case where a NumPy `ndarray` was created with USM. For such cases, the check should be if the `arr.base` has `__sycl_usm_array_interface__` defined.

PokhodenkoSA:
Closes #41 
Closes #33

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -23,6 +23,7 @@ requirements:
     run:
         - python
         - numba >=0.51
+        - cffi
         - dpctl
         - spirv-tools
         - llvm-spirv
--- a/numba_dppy/compiler.py
+++ b/numba_dppy/compiler.py
@@ -11,7 +11,7 @@ from types import FunctionType
 from inspect import signature
 
 import dpctl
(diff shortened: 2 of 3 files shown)


PokhodenkoSA:
CI passed. I will merge is because it is blocker now.
@diptorupd @reazulhoque Please, review it later.

Issue #39, opened 2020-12-02
NUMBA_OPT value has no impact on generated IR


akharche:
Changing the NUMBA_OPT environment variable has no impact on generated IR. 
It was expected the IR would show at least some changes when NUMBA_OPT is changed from 0 to 3.

akharche:
Numba-dppy fails when changing to the "-O0" spirv generation

Numba-dppy ("-O0") and dpcpp generated LLVM IRs
[numba_dppy_llvm_O0.log](https://github.com/IntelPython/numba-dppy/files/5814993/numba_dppy_llvm_O0.log)
[dpcpp_kernel_sum_llvm.log](https://github.com/IntelPython/numba-dppy/files/5814994/dpcpp_kernel_sum_llvm.log)

PR #62, opened 2020-12-10
Enable optimization level setting

akharche:
Issue #39


PokhodenkoSA (changes requested):
Please, merge with master.

akharche:
This PR breaks GDB support when -O0 optimization level is set

akharche:
@reazulhoque @diptorupd  If we take `llvm-spirv` from oneapi compiler, it solves the problem with crash on "-O0" opt level

diptorupd (approved):

numba_dppy/codegen.py
```diff
@@ -38,9 +39,14 @@ def _optimize_functions(self, ll_module):
...
 
     def _optimize_final_module(self):
+        opt_level_option = config.OPT
+
         # Run some lightweight optimization to simplify the module.
         pmb = ll.PassManagerBuilder()
-        pmb.opt_level = 1
+
+        # Turn off any optimizations if OPT level is set to 0
+        pmb.opt_level = 0 if opt_level_option == 0 else 1
```
PokhodenkoSA, line 48: Why `pmb.opt_level` is not equal to `config.OPT`?

reazulhoque:
@PokhodenkoSA @diptorupd should we merge this?

akharche:
@reazulhoque I think, it should be merge after problem with -O0 (addr space casting) is fixed. This code will fail if user changes option NUMBA_OPT to 0

PokhodenkoSA:
> This code will fail if user changes option NUMBA_OPT to 0

It is up to user to use any NUMBA_OPT value. This PR only provides an option to configure.
If default NUMBA_OPT value is appropriate we could merge this PR.

PokhodenkoSA (approved):

PR #63, opened 2020-12-10
Rewrite supported ndarray.function_name

reazulhoque:
This PR adds a new typed pass to rewrite `ndarray.function_name` (e.g. a.sum()) to `numba_dppy.dpnp.function_name(ndarray)` (e.g. numba_dppy.dpnp.sum(a)).

We need a typed pass to make sure the array is in fact an ndarray. 

Test for the pass and ndarray.sum() is added.

The rewrite pass was mostly borrowed from canonicalize_array_math() defined in numba/core/ir_utils.py from Numba.


PokhodenkoSA:
@reazulhoque Please, analyze tests fails on CI.

reazulhoque:
Tested with the following dependencies locally:

1. dpctl: d157ee0b1a9ba88d0f3b226bd069af2dd9a56b9a
2. Numba (patched): 864790061aa7548ac2b7ebcf448354def849e657
3. dpnp: 0.4.0

Result:
```
----------------------------------------------------------------------
Ran 148 tests in 101.530s 
OK (skipped=2, expected failures=1)
```

PokhodenkoSA (approved):
LGTM. CI passed.

numba_dppy/tests/test_rename_numpy_function_pass.py
```diff
@@ -75,6 +76,7 @@ def got(a):
         self.assertTrue(check_equivalent(expected_ir, pipeline.state.func_ir))
 
 
+@unittest.skipUnless(ensure_dpnp(), 'test only when dpNP is available')
```
reazulhoque, line 79: We need to test the rewrites regardless of whether dpnp is available. This is so because we enable this functions through dpnp and we need to rewrite them in order to fail if dpnp is available.

PR #118, opened 2020-12-15
Spokhode/dparray tests

PokhodenkoSA:
Convert examples style tests to unittest.

### DIFF
--- a/numba_dppy/tests/test_usmarray.py
+++ b/numba_dppy/tests/test_usmarray.py
@@ -1,6 +1,7 @@
 import numba
 import numpy
 import sys
+import unittest
 
 from numba_dppy.testing import DPPYTestCase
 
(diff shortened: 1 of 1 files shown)


numba_dppy/tests/test_usmarray.py
```diff
@@ -1,228 +1,204 @@
-from __future__ import print_function, division, absolute_import
-
 import numba
-import dpctl.dptensor.numpy_usm_shared as usmarray
 import numpy
 import sys
+import unittest
 
+from numba_dppy.testing import DPPYTestCase
```
1e-to: We don't use DPPYTestCase anymore.
Use unittest.TestCase instead

DrTodd13 (commented):
Generally looks fine with my previous comment that the USM to numpy implicit conversion test should be removed.

PR #120, opened 2020-12-15
Skip tests

reazulhoque:
This PR skips certain tests on some architecture.

PR #115, opened 2020-12-15
Port existing functions enabled through dpnp

reazulhoque:
This PR ports the implementation of functions through dpnp using the new approach. It cleans up the older approach as well.

PR #125, opened 2020-12-16
Port existing dpnp functions

1e-to:


reazulhoque:
Is this a diplicate of https://github.com/IntelPython/numba-dppy/pull/115? Why do we need this? @PokhodenkoSA do you know why this PR is passing the CI but 115 is not?

PokhodenkoSA:
> Is this a diplicate of #115? Why do we need this?

@1e-to take #120 as we discussed.
I think this branch was created because I say that it is impossible to push commits into your branch. But I was wrong, now I think it is possible and this PR is not necessary. @1e-to if you have commits for @reazulhoque's PR please push to his repo in the branch from original PR.

PR #190, opened 2021-01-21
[WIP] Feature - Convert all input of dpnp functions to be backed by USM

reazulhoque:
This PR inserts a function call before any dpnp functions to convert all the arguments into a USM backed argument if it is not already the case.


diptorupd:
The ongoing work towards support for compute follows data logic in dpex njit will remove all support for ndarrays in the auto-offload portion of the pipeline. Thus, there is no point in reviving this work.

Issue #203, opened 2021-01-25
Example for RNG on GPU


PokhodenkoSA:
Provide example for using RNG on GPU in jitted region. Show integration with dpNP.

Random is provided by https://github.com/IntelPython/numba-dppy/pull/141.

PR #215, opened 2021-01-28
Add example for rng

akharche:
Issue #203

### DIFF
--- a/numba_dppy/examples/rand.py
+++ b/numba_dppy/examples/rand.py
@@ -24,7 +24,7 @@ def random_normal(loc, scale, size):
 
 
 size = 9
-scale = 3.
+scale = 3.0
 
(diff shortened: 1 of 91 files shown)


numba_dppy/examples/rand.py
```diff
@@ -0,0 +1,44 @@
...
+    return np.random.exponential(scale, size)
+
+
+@numba.njit
+def random_normal(loc, scale, size):
+    return np.random.normal(loc, scale, size)
+
+
+size = 9
+scale = 3.
```
PokhodenkoSA, line 27: I think black formatter will try to format this to `3.0`.

PokhodenkoSA:
How to check that functions are executed on GPU?

PokhodenkoSA (approved):

PR #231, opened 2021-02-02
Add documentation about supported NumPy functions via DPNP

reazulhoque:
This PR uses jinja to auto generate a stub class representing the NumPy functions we currently support. This generation also produces a generic docstring for each function that can help us in documenting which NumPy functions we support.


diptorupd:
@reazulhoque what's the status on this PR? Can it be added to the docs for this release?

reazulhoque:
No, this intriduces disruptive changes to how we implement function from dpnp in Numba_dppy and will be risky to add now.

PokhodenkoSA (commented):
It is very old PR but it is actual now because @samaid asks for clear understanding what NumPy features numba-dppy support.

diptorupd:
Too bit rotten at this point.

PR #234, opened 2021-02-03
Implement np.sort in kernel by dpnp

1e-to:
Closes #225

PR #246, opened 2021-02-10
Implement np.take in kernel by dpnp

vlad-perevezentsev:


numba_dppy/tests/test_dpnp_functions.py
```diff
@@ -1411,5 +1411,88 @@ def f(a, b):
...
+            return c
+
+        test_indices = []
+        test_indices.append(np.array([1, 5, 1, 11, 3]))
+
+        with assert_dpnp_implementaion():
+            for ind in test_indices:
+                self.assertTrue(
+                    self.check_take_for_different_datatypes(
+                        f, np.take, self.get_np_array(ind), [12], self.tys, matrix=False
```
akharche, line 1464: I think you can just cast value to np.array here for indexes. Function `get_np_array` is redundant

akharche:
@reazulhoque Could you please look into this PR?

numba_dppy/tests/test_dpnp_functions.py
```diff
@@ -1411,5 +1411,88 @@ def f(a, b):
...
+            raise TypeError
+
+    def test_take_types(self):
+        @njit
+        def f(a, ind):
+            c = np.take(a, ind)
+            return c
+
+        test_indices = []
+        test_indices.append(np.array([1, 5, 1, 11, 3]))
```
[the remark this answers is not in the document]
reazulhoque: Agree with Angelina.

numba_dppy/tests/test_dpnp_functions.py
reazulhoque: I also think the get_np_array is redundant, what are we trying to test throught this function?

numba_dppy/tests/test_dpnp_functions.py
```diff
@@ -1411,5 +1411,84 @@ def f(a, b):
...
+        with assert_dpnp_implementaion():
+            self.assertTrue(
+                self.check_take_for_different_datatypes(
+                    f, np.take, np.array([1, 5, 1, 11, 3]), [12], self.tys, matrix=False
+                )
+            )
+
+    def test_take_indices(self):
+        @njit
+        def f(a, ind):
```
reazulhoque, line 1466: Both the `f()` function seems to perform the same thing, may be 1 common is good enough. Line 1453 and 1466.

numba_dppy/tests/test_dpnp_functions.py
```diff
@@ -1411,5 +1411,84 @@ def f(a, b):
...
+)
+class Testdpnp_array_ops_functions(unittest.TestCase):
+    tys = [np.int32, np.uint32, np.int64, np.uint64, np.float, np.double]
+
+    def check_take_for_different_datatypes(
+        self, fn, test_fn, ind, dims, tys, matrix=None
+    ):
+        for ty in tys:
+            if matrix:
+                a = np.arange(np.prod(np.array(dims)), dtype=ty).reshape(
```
reazulhoque, line 1425: If `dims` is expected to be a `np.array` may be let us pass an `np.arra`y instead of a `list`.

numba_dppy/tests/test_dpnp_functions.py
```diff
@@ -1411,5 +1411,84 @@ def f(a, b):
...
 
 
+@unittest.skipUnless(
+    ensure_dpnp() and dpctl.has_gpu_queues(), "test only when dpNP and GPU is available"
+)
+class Testdpnp_array_ops_functions(unittest.TestCase):
+    tys = [np.int32, np.uint32, np.int64, np.uint64, np.float, np.double]
+
+    def check_take_for_different_datatypes(
+        self, fn, test_fn, ind, dims, tys, matrix=None
```
reazulhoque, line 1421: Since `matrix` seems to be a boolean, let us assign a boolean as a default value?

reazulhoque:
@vlad-perevezentsev  This should be merged with main and also Elena has a [PR](https://github.com/IntelPython/numba-dppy/pull/234) that adds the same file and other function of same category, we need to be aware of that.

numba_dppy/tests/test_dpnp_functions.py
vlad-perevezentsev: @reazulhoque 
The numpy.take function can accept indexes of the list and tuple types. We need to get the shape from the indexes to set the shape of the out array.
The get_np_array function allows us to pass indexes of the list and tuple types.

PokhodenkoSA (approved):
LGTM

PR #248, opened 2021-02-11
Don't create a new copy of a usm shared array data pointers for kernel call.

DrTodd13:
In the code that forms the kernel args, we now look to see if the array type is actually a USM shared type.  If so, we just place the data pointer directly into the kernel arg and so we avoid all the overhead of creating a new USM buffer and copying data there before the kernel and back afterwards.

### DIFF
--- a/numba_dppy/dppy_host_fn_call_gen.py
+++ b/numba_dppy/dppy_host_fn_call_gen.py
@@ -22,6 +22,7 @@ import llvmlite.binding as lb
 from numba.core import types, cgutils
 
 from numba.core.ir_utils import legalize_names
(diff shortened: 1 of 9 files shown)


reazulhoque (commented):
@DrTodd13  This branch needs to merge with main. LGTM after that.

PokhodenkoSA:
Waiting for test cases from @reazulhoque.

reazulhoque:
@PokhodenkoSA This PR is ready for review.

akharche:
test_no_copy_usm_shared faild on Windows

reazulhoque (approved):

PR #263, opened 2021-02-24
Add np.copy via dpnp

1e-to:


numba_dppy/rename_numpy_functions_pass.py
```diff
@@ -91,6 +91,7 @@
     "cumsum": (["numpy"], "cumsum"),
     "cumprod": (["numpy"], "cumprod"),
     "sort": (["numpy"], "sort"),
+    "copy": (["numpy"], "copy"),
```
PokhodenkoSA, line 94: Please, sort lines in ABC

PokhodenkoSA (changes requested):

PokhodenkoSA (approved):
LGTM

PR #275, opened 2021-03-03
Automate release versions in docs

vlad-perevezentsev:


PokhodenkoSA (approved):

PokhodenkoSA (changes requested):
The next version will be 0.13.1.
Could you find a way how to not hardcode version but get it from current sources version?

PR #287, opened 2021-03-15
Introduce DPPYArray type and DPPYArrayModel

reazulhoque:
This PR is ready to be merged but will work after https://github.com/numba/numba/pull/6469 is merged in Numba master. I put the WIP tag to stop merging until that happens.

Changes reflecting https://github.com/numba/numba/pull/6469 was merged in Intel Numba(18431c019e764b986f59974f4d813c41662f2146) in branch https://github.com/IntelPython/numba/commits/release0.53-patched. Once this branch gets updated in CI we will need to merge this in Numba_dppy.


diptorupd:
@reazulhoque Would it be possible to overload the `from_argument` function in our data model to insert an address space cast for any pointer that has a global address space.

numba_dppy/dppy_array_type.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def is_precise(self):
+        return self.dtype.is_precise()
+
+
+class DPPYArrayModel(StructModel):
+    def __init__(self, dmm, fe_type):
+        ndim = fe_type.ndim
+        members = [
+            ("meminfo", types.MemInfoPointer(fe_type.dtype)),
```
diptorupd, line 87: Couple of questions @reazulhoque @DrTodd13 @PokhodenkoSA:

1. Do we need the `meminfo` and `parent` attributes? We are not using it in any place?
2. Will it be possible to add an address space specifier to the two attributes based on our `GenericPointerModel`?

It will be useful for my work on removing the kernel wrapper function. I made the needed changes so that address space casting of kernel arguments is done during the `decode_arguments` phase at the time of lowering. However, the next step of unflattening the array arguments into a struct fails because the data model does not match.

numba_dppy/ocl/ocldecl.py
```diff
@@ -164,7 +165,7 @@ def typer(shape, dtype):
             ndim = parse_shape(shape)
             nb_dtype = parse_dtype(dtype)
             if nb_dtype is not None and ndim is not None:
-                return types.Array(
+                return DPPYArray(
                     dtype=nb_dtype,
                     ndim=ndim,
                     layout="C",
```
diptorupd, line 171: Do we envision supporting strides and non-C contiguous data layouts?

PokhodenkoSA: Moved to https://github.com/IntelPython/numba-dppy/issues/363.

numba_dppy/dppy_array_type.py
PokhodenkoSA: Moved to https://github.com/IntelPython/numba-dppy/issues/364.
Could be implemented in separate PR.

PR #291, opened 2021-03-23
Fix eager compilation with signature for dppy.kernel

### LINKED ISSUE #75: Make sure consistent Sycl queue is being used in eager compilation case
In case when we are compiling kernel ahead of time, we need to make sure the sycl queue being used are for the same type of backend. Right now, the get_current_queue() will return the default queue which might be using a backend that does not match the backend of the context users will be calling the kernel from. We need to check at call time if the backend matches if not we need to recmopile and the cached kernel can not be used.

For e.g. the default queue of dpctl is set to `"level0:gpu"` and the following program will execute correctly:
```
import numpy as np
import numba_dppy as dppy
import dpctl

filter_str = "level0:gpu"

@dppy.kernel("void(float32,float32,float32)")
def data_parallel_sum(a, b, c):
    c = a + b


def main():
    global_size = 10
    N = global_size
    print("N", N)

    a = np.array(np.random.random(N), dtype=np.float32)
    b = np.array(np.random.random(N), dtype=np.float32)
    c = np.ones_like(a)

    with dpctl.device_context(filter_str) as gpu_queue:
        data_parallel_sum[global_size, dppy.DEFAULT_LOCAL_SIZE](a[0], b[0], c[0])

    print("Done...")


if __name__ == "__main__":
    main()
```

OUTPUT:
```
N 10
Done...
```

if we change the `filter_str` to `"opencl:gpu"`, here is the output we get:
```
N 10
Segmentation fault (core dumped)
```

reazulhoque:
Closes #75 

This will fix the seg-fault issue when signature was provided with `dppy.kernel`. 

Eager compilation is still Just-In-Time compilation. Instead of the compilation happening at the call site it happens at the definition site. We have caching mechanism for already compiled kernels that takes the SYCL queue in account. This
 mechanism can be potentially used to recompile a kernel that has been compiled for a different backend (Sycl queue).

Steps taken to fix this:

- Make sure when we provide signature, that triggers the eager compilation, we go through the existing caching mechanism. If the Sycl queue is consistent we simply launch the kernel in other cases we recompile for the given queue.


PokhodenkoSA (changes requested):
Please fix CI on Windows.

PokhodenkoSA (changes requested):
Can we test AOT somehow?

PokhodenkoSA:
@reazulhoque please merge with main. It will fix black.

numba_dppy/tests/kernel_tests/test_barrier.py
```diff
@@ -81,18 +91,22 @@ def test_local_memory(filter_str):
...
+
+            # preload
+            lm[i] = A[i]
+            # barrier local or global will both work as we only have one work group
+            dppy.barrier(dppy.CLK_LOCAL_MEM_FENCE)  # local mem fence
+            # write
+            A[i] += lm[blocksize - 1 - i]
+
+    except:
+        pytest.skip()
```
PokhodenkoSA, line 109: It is strange that if it does not work then it is skipped. Why it could not work?

PokhodenkoSA: Tests should pass or fail. They should not skip on fail. If we expect some situation incompatible with test then we should test it preliminary as a prerequisite.

PokhodenkoSA (changes requested):
I am continuing to review.

numba_dppy/tests/kernel_tests/test_barrier.py
reazulhoque: @PokhodenkoSA it could not work because `dpctl` does not support level0 yet on Windows.

Can you please suggest an example of the prerequsite? I am thinking of testing if level0 bakcend is supported when the platform is Windows and depending on that skip tests if it is not supported. But in that case we again end up skipping on fail of the prerequisite. In that case the good thing is the failure is known and in the current case it is not. Please share your thoughts.

PokhodenkoSA: > Can you please suggest an example of the prerequisite?

Before running test check that Level Zero is supported.

> But in that case we again end up skipping on fail of the prerequisite.

When you test expected state of the system like checking that Level Zero is available before running test code you explicitly show to reader of the code that it is necessary to have Level Zero for the code in test to work.
Otherwise user could think that exceptions in this code is expected and normal behavior. So test does not show some stable results users can rely on.

numba_dppy/tests/kernel_tests/test_barrier.py
```diff
@@ -91,23 +90,21 @@ def test_local_memory(filter_str):
...
-
-            # preload
-            lm[i] = A[i]
-            # barrier local or global will both work as we only have one work group
-            dppy.barrier(dppy.CLK_LOCAL_MEM_FENCE)  # local mem fence
-            # write
-            A[i] += lm[blocksize - 1 - i]
-
-    except:
+    if sys.platform in ["win32", "cygwin"] and filter_str == list_of_filter_strs[1]:
```
PokhodenkoSA: pytest has `skipif` test decorator (https://docs.pytest.org/en/reorganize-docs/new-docs/user/skipping.html).
Moreover, it provides example for reusing this skips (https://docs.pytest.org/en/reorganize-docs/new-docs/user/skipping.html#re-using-skipif-decorators).
Reusing `skippers` eliminates code duplication and makes code more readable.

PokhodenkoSA (changes requested):
Much better. I have made remarks with useful links describing how to make better.

numba_dppy/decorators.py
```diff
@@ -21,6 +20,7 @@
     compile_dppy_func,
     get_ordered_arg_access_types,
 )
+import dpctl
```
PokhodenkoSA, line 23: Wrong import order.

PokhodenkoSA:
This PR does not fix https://github.com/IntelPython/numba-dppy/issues/75. Because the test uses `func.specialize(..., gpu_queue)` and the issue have another scenario.

numba_dppy/compiler.py
```diff
@@ -545,32 +545,39 @@ def check_for_invalid_access_type(self, access_type):
...
 
 class JitDPPYKernel(DPPYKernelBase):
-    def __init__(self, func, access_types):
+    def __init__(self, func, debug, access_types):
 
         super(JitDPPYKernel, self).__init__()
 
         self.py_func = func
         self.definitions = {}
+        self.debug = debug
```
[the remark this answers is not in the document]
reazulhoque: This variable is to forward the debug flag provided in the `@dppy.kernel` decorator. We are not using it right now but we allow users to provide the debug flag in the decorator. I am saving it so that in the future when we decide to implement the desired functionality we can know when users have provided it later in the pipeline.

numba_dppy/decorators.py
reazulhoque: Fixed.

numba_dppy/tests/kernel_tests/test_barrier.py
```diff
@@ -31,11 +32,28 @@ def filter_str(request):
     return request.param
 
 
+def skip_if_win():
+    if sys.platform in ["win32", "cygwin"]:
```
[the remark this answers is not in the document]
reazulhoque: The question directs to an already answered question which suggests `sys.platform`.

reazulhoque:
@PokhodenkoSA, this PR solves https://github.com/IntelPython/numba-dppy/issues/75. There are multiple tests that needed to be updated, which test are you talking about when you mention `func.specialize(..., gpu_queue)`? 

I have updated tests with signatures that will trigger eager compilation. I asked Numba team specifically if they have tests for eager compilation and they mentioned they do not. I decided to not write specific tests following that conversation.

When I add signature to `@dppy.kernel` in exisiting test it follows the example I have provided in #75. Tests I have updated to make sure eager compilation is working:
1. https://github.com/IntelPython/numba-dppy/pull/291/files#diff-b7c075c42e2cbadf69e707181714af75c0c8a705a806e2e2f71345253fc7af73R56
2. https://github.com/IntelPython/numba-dppy/pull/291/files#diff-b7c075c42e2cbadf69e707181714af75c0c8a705a806e2e2f71345253fc7af73R81
3. https://github.com/IntelPython/numba-dppy/pull/291/files#diff-b7c075c42e2cbadf69e707181714af75c0c8a705a806e2e2f71345253fc7af73R108
4.

numba_dppy/compiler.py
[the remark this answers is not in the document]
PokhodenkoSA: Ok

numba_dppy/tests/kernel_tests/test_barrier.py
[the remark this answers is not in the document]
PokhodenkoSA: `platform.system() == "Windows"`

PokhodenkoSA:
I agree with test changes. Need CI fix and use `platform.system()`.

numba_dppy/compiler.py
```diff
@@ -561,32 +561,42 @@ def check_for_invalid_access_type(self, access_type):
...
         self.access_types = access_types
 
         from .descriptor import dppy_target
 
         self.typingctx = dppy_target.typing_context
 
+    def get_argtypes(self, *args):
```
diptorupd, line 577: Please use `_get_argtypes` if the function is not public API.

PokhodenkoSA (approved):

PR #297, opened 2021-03-29
Initial tests for debug info

akharche:


numba_dppy/tests/test_debug.py
```diff
@@ -0,0 +1,91 @@
...
+# limitations under the License.
+
+import re
+import pytest
+
+import numba_dppy as dppy
+import dpctl
+
+from numba.core import types, compiler
+from numba_dppy import compiler
```
PokhodenkoSA, line 23: `compiler` imported twice.

numba_dppy/tests/test_debug.py
```diff
@@ -0,0 +1,91 @@
...
+def offload_device(request):
+    return request.param
+
+
+def get_kernel_ir(fn, sig, debug=False):
+    kernel = compiler.compile_kernel(fn.sycl_queue, fn.py_func, sig, None, debug=debug)
+    return kernel.assembly
+
+
+def make_check(fn, sig, expect):
```
PokhodenkoSA, line 44: This function makes a lot:
1. compile (with signature and debug options)
2. use regexp to search for debug symbols
3. make an assertion
To much responsibilities for the function. Also when it is called from test code it usually requires to see inside it to understand what exactly it does.

numba_dppy/tests/test_debug.py
```diff
@@ -0,0 +1,91 @@
...
+
+    debug_expect = False
+
+    @dppy.kernel
+    def foo(x):
+        return x
+
+    with dpctl.device_context(offload_device):
+        sig = (types.int32,)
+        make_check(foo, sig, debug_expect)
```
PokhodenkoSA, line 91: Both tests are the same. Only difference `debug_expect` is True or False. Could you please eliminate duplication and make `debug_expect` as test parameter (https://docs.pytest.org/en/stable/parametrize.html)?

PokhodenkoSA (changes requested):
Please fix comments.

numba_dppy/tests/test_debug.py
PokhodenkoSA: flake8 in CI will find issues with imports.

numba_dppy/tests/test_debug.py
akharche: I agree, I did it on purpose to clearly determine and split tests that responsible for different functionality. If I provide a fixture it may be not clear which test is failed. I mean this implementation makes a distinction by test name.

numba_dppy/tests/test_debug.py
akharche: It's just incapsulated to this function. Ok, I can move 1 and 3 to test

numba_dppy/tests/test_debug.py
PokhodenkoSA: Test could check that `debug` flag works in both True and False states.

numba_dppy/tests/test_debug.py
```diff
@@ -0,0 +1,81 @@
...
+
+import re
+import pytest
+
+import numba_dppy as dppy
+import dpctl
+
+from numba.core import types
+from numba_dppy import compiler
+from numba_dppy.tests.skip_tests import skip_test
```
PokhodenkoSA, line 24: pylint shows following warnings:
```bash
numba_dppy/tests/test_debug.py:20:0: C0411: third party import "import dpctl" should be placed before "import numba_dppy as dppy" (wrong-import-order)
numba_dppy/tests/test_debug.py:22:0: C0411: third party import "from numba.core import types" should be placed before "import numba_dppy as dppy" (wrong-import-order)
```
I know that until CI will perform code style checks this kind of things will be common.
I will make this checks in CI soon. For now I can only recommend start pylint locally.

PokhodenkoSA (approved):
Much better. Please, fix my remarks if you want.

akharche:
To make fixtures visible for all the tests, moved them to separate config file.
To know more about pytest features look to docs:
https://docs.pytest.org/en/stable/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session

akharche:
@reazulhoque For some reason tests on "level0" are failed here. Error is "SyclProgramCompilationError", so I temporarily disabled them. Could you please look into get_kernel_ir() function in the test, may be I do something wrong for level0?

PR #299, opened 2021-03-29
[WIP] dppy.kernel queue keyword

reazulhoque:
This PR introduces passing a queue through the `@dppy.kernel` decorator.


PokhodenkoSA (changes requested):
1. One test is not enough. 
2. Need docs updating for explaining semantics.
3. Tests need to be updated to new style with using `filter_strings` parameters.

diptorupd:
The current design to move towards compute follows data makes these changes deprecated.

Issue #302, opened 2021-03-31
Lowering multiple pranges are failing for GPU


reazulhoque:
Running the following program produces:

```
import numpy as np
import numba
import dpctl
from numba import njit, prange
import numba_dppy

def test_nested_prange():
    @njit
    def f(a, b):
        # dimensions must be provided as scalar
        m, n = a.shape
        for i in prange(m):
            for j in prange(n):
                b[i, j] = a[i, j] * 10

    m = 8
    n = 8
    a = np.ones((m, n))
    b = np.ones((m, n))

    with dpctl.device_context("opencl:gpu"):
        f(a, b)

test_nested_prange()
```

Output:
```
To see details set environment variable NUMBA_DPPY_DEBUG=1

File "test_multiple_prange.py", line 17:
    def f(a, b):
        <source elided>
        m, n = a.shape
        for i in prange(m):
        ^

  warnings.warn(NumbaPerformanceWarning(msg, parfor.loc))
OMP: Info #276: omp_set_nested routine deprecated, please use omp_set_max_active_levels instead.
```

We need to investigate and fix this issue.

PR #303, opened 2021-03-31
Remove dead code only if function name is replaced

reazulhoque:
This PR makes sure we try to remove dead code only when we replace the Numpy function names with our own. Removing dead code regardless was failing in cases where we have multiple `prange` as mentioned in [this issue](https://github.com/IntelPython/numba-dppy/issues/302). This PR should close https://github.com/IntelPython/numba-dppy/issues/302.


PokhodenkoSA (changes requested):
Please, provide test case.

reazulhoque:
@PokhodenkoSA this bug was noticed originally because [this test](https://github.com/IntelPython/numba-dppy/blob/main/numba_dppy/tests/test_prange.py) was falling back. This PR fixes this test from falling back and offloads it through our pipeline. Should we ask the original author of that test to make sure it fails if it is falling back instead of going through our pipeline?

I can always write tests and that would require setting up a partial compiler pipeline to test it, which is tedious. Please share your thoughts.

PokhodenkoSA:
> [this test](https://github.com/IntelPython/numba-dppy/blob/main/numba_dppy/tests/test_prange.py) was falling back.

Does it mean that test did not check that code falling back? So test was blind?
No when you fix the code did you fix the test to check that fallback is not happening now?
I think test should be fixed first to catch bad situation and show that the test fails. Then your fix for the feature should show that it makes test passing.

reazulhoque:
>> Does it mean that test did not check that code falling back? So test was blind?
Yes.

I have updated the tests to check for falling back.

numba_dppy/tests/test_prange.py
```diff
@@ -81,9 +88,13 @@ def f(a, b):
...
         b = np.ones((m, n))
 
-        with dpctl.device_context("opencl:gpu"):
+        old_debug = numba_dppy.compiler.DEBUG
+        numba_dppy.compiler.DEBUG = 1
+
+        with captured_stdout() as stdout, dpctl.device_context("opencl:gpu"):
             f(a, b)
 
+        self.assertTrue("Parfor lowered to specified SYCL device" in stdout.getvalue())
```
PokhodenkoSA, line 97:
```suggestion
        self.assertIn("Parfor lowered to specified SYCL device", stdout.getvalue())
```

numba_dppy/tests/test_prange.py
```diff
@@ -135,12 +150,14 @@ def prange_example():
         numba_dppy.compiler.DEBUG = old_debug
 
         self.assertEqual(
-            stdout.getvalue().count("Parfor lowered on DPPY-device"),
+            stdout.getvalue().count(""),
```
PokhodenkoSA, line 153:
```suggestion
            stdout.getvalue().count("Parfor lowered to specified SYCL device"),
```

numba_dppy/tests/test_prange.py
```diff
@@ -37,7 +33,7 @@ def f(a, b):
         a = np.ones((m, n))
         b = np.ones((m, n))
 
-        with dpctl.device_context("opencl:gpu"):
+        with assert_auto_offloading(1, 0), dpctl.device_context("opencl:gpu"):
```
PokhodenkoSA, line 36:
```suggestion
        with assert_auto_offloading(parfors=1, 0), dpctl.device_context("opencl:gpu"):
```

numba_dppy/testing.py
```diff
@@ -76,3 +77,36 @@ def assert_dpnp_implementaion():
         yield
 
     assert "dpnp implementation" in stdout.getvalue(), "dpnp implementation is not used"
+
+
+@contextlib.contextmanager
+def assert_auto_offloading(
+    expected_offloaded_count=-1, expected_not_offloaded_count=-1
```
PokhodenkoSA, line 84:
```suggestion
    parfors=-1, expected_not_offloaded_count=-1
```

PokhodenkoSA, line 84: We could find better name for parameters.

numba_dppy/tests/test_prange.py
PokhodenkoSA, line 36: Use keyword arguments.

numba_dppy/testing.py
PokhodenkoSA: Use 0 as default value for the second parameter.

numba_dppy/testing.py
```diff
@@ -76,3 +77,36 @@ def assert_dpnp_implementaion():
...
+):
+    old_debug = numba_dppy.compiler.DEBUG
+    numba_dppy.compiler.DEBUG = 1
+
+    with captured_stdout() as stdout:
+        yield
+
```
PokhodenkoSA, line 91: Reset DEBUG to old value.

reazulhoque:
@PokhodenkoSA @diptorupd Can you please review it?

PokhodenkoSA (approved):

PR #309, opened 2021-04-04
Change all Numba-dppy examples to work wih dpctl 0.7.0

diptorupd:
Update all examples to use the latest dpctl 0.7.0 API pushed to main in #340.
Other improvements to examples.


diptorupd:
There are few pending issues that need to be looked at:

- `examples/blackscholes_njit.py` segfaults when used with `level zero`.
- `tests/test_rename_numpy_function_pass.py:91` fails (may not be related to this changes)

I have not tested dpnp tests and examples.

diptorupd:
@reazulhoque  I investigated the `examples/blackscholes_njit.py` segfaults and it might be an issue that already existed and is just getting caught with by the changes I made to the example. The issue seems to be that either we are reusing the cached program that was compiled for OpenCL on level zero or we are passing an OpenCL allocated USM pointer to level zero and it is crashing inside DPC++. I do not know if there is anything we can do inside dpctl to prevent the crash, we need to get to the root cause before that.

reazulhoque:
The issue in fact seems to be that we re relying on `njit`'s caching mechanism which does not cascade down to the kernel level. So, we are reusing the `njit` function once we compile it even when the kernel we are generating for the `parfor` regions are not the same.

diptorupd:
> The issue in fact seems to be that we re relying on `njit`'s caching mechanism which does not cascade down to the kernel level. So, we are reusing the `njit` function once we compile it even when the kernel that should be generated for the `parfor` regions are not the same.

As discussed, I have changed the failing example and opened an issue ticket #343 to track the caching issue.

numba_dppy/examples/atomic_op.py
```diff
@@ -13,25 +13,34 @@
...
+    try:
+        d = dpctl.select_gpu_device()
+        with dpctl.device_context(d):
+            print("Offloading to ...")
+            d.print_device_info()
+            atomic_add[global_size, dppy.DEFAULT_LOCAL_SIZE](a)
+            # Expected 100, because global_size = 100
+            print(a)
+    except ValueError:
+        print("No OpenCL GPU found.")
```
reazulhoque, line 43: How do we know the device is of OpenCL type in this case?

numba_dppy/examples/sum_reduction_recursive_ocl.py
```diff
@@ -79,13 +83,16 @@ def sum_recursive_reduction(size, group_size, Dinp, Dpartial_sums):
...
 
 
-def get_context():
-    if dpctl.has_gpu_queues():
-        return "opencl:gpu"
-    elif dpctl.has_cpu_queues():
-        return "opencl:cpu"
-    else:
-        raise RuntimeError("No device found")
+def get_device():
```
reazulhoque, line 86: We should move this function inside _helper.py

numba_dppy/examples/atomic_op.py
diptorupd: done.

numba_dppy/examples/sum_reduction_recursive_ocl.py
diptorupd: done.

PR #310, opened 2021-04-04
Use correct names for other products dpctl, Numba, dpnp

diptorupd:
using lower case for `dppy`, `dpnp`, `dpctl`. The style of camel casing just did not work with the masses.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ https://intelpython.github.io/dpnp/
 
 ## dppy
 
-dppy is a proof-of-concept backend for NUMBA to support compilation for
+dppy is a proof-of-concept backend for Numba to support compilation for
 Intel CPU and GPU architectures.
 The present implementation of dpPy is based on OpenCL 2.1, but is likely
 to change in the future to rely on Sycl/DPC++ or Intel Level-0 driver API.
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # numba-dppy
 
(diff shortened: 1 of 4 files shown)


PokhodenkoSA:
> The style of camel casing just did not work with the masses.

True. Let's do it in all docs too.

README.md
```diff
@@ -22,16 +22,16 @@ https://intelpython.github.io/dpnp/
...
 * llvm-spirv (SPIRV generation from LLVM IR)
 * llvmdev (LLVM IR generation)
 * spirv-tools
 * scipy (for testing)
 
-## dpPy
+## dppy
 
-dpPy is a proof-of-concept backend for NUMBA to support compilation for
+dppy is a proof-of-concept backend for NUMBA to support compilation for
```
PokhodenkoSA, line 34:
```suggestion
dppy is a proof-of-concept backend for Numba to support compilation for
```

PokhodenkoSA: We also should name Numba consistently. The name of product is `Numba`, but package name is `numba`.

PokhodenkoSA (approved):
Good to start.

README.md
diptorupd: Done. I have fixed few more places in docs. More might remain.

PokhodenkoSA: Please continue. Do not what is best with Numba. Possibly most cases should be `Numba`, but if we posint to pakcage name then `numba`.

PR #316, opened 2021-04-13
Add DeviceArray class

reazulhoque:
This PR introduces `DPPYDeviceArray` class in Numba_dppy. This will give users: 

-  `class` to represent arrays in device.
- Will allow Numba_dppy to implement interface Numba requires for functionality such as `@vectorize`. See [here](https://github.com/numba/numba/blob/master/numba/np/ufunc/deviceufunc.py#L308).

### DIFF
--- a/numba_dppy/compiler.py
+++ b/numba_dppy/compiler.py
@@ -406,9 +406,9 @@ class DPPYKernel(DPPYKernelBase):
         retr = []  # hold functors for writeback
         kernelargs = []
         internal_device_arrs = []
-        for ty, val, access_type in zip(
+        for idx, (ty, val, access_type) in enumerate(zip(
             self.argument_types, args, self.ordered_arg_access_types
-        ):
+        )):
             self._unpack_argument(
                 ty,
                 val,
@@ -417,6 +417,7 @@ class DPPYKernel(DPPYKernelBase):
                 kernelargs,
                 internal_device_arrs,
                 access_type,
+                idx
             )
 
         self.sycl_queue.submit(
@@ -468,7 +469,7 @@ class DPPYKernel(DPPYKernelBase):
             kernelargs.append(ctypes.c_longlong(val.strides[ax]))
 
     def _unpack_argument(
-        self, ty, val, sycl_queue, retr, kernelargs, device_arrs, access_type
+        self, ty, val, sycl_queue, retr, kernelargs, device_arrs, access_type, idx
     ):
         """
         Convert arguments to ctypes and append to kernelargs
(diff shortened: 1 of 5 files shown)


numba_dppy/dppy_devicearray.py
```diff
@@ -0,0 +1,148 @@
...
+# limitations under the License.
+
+import numpy as np
+import dpctl.memory as dpctl_mem
+import dpctl
+from numba.core import types
+from numba.np import numpy_support
+
+
+class DPPYDeviceArray(object):
```
PokhodenkoSA, line 22: Is is necessary to name it with DPPY prefix? It is from dppy package already.

PokhodenkoSA:
@reazulhoque please describe what this class should do?

reazulhoque:
@PokhodenkoSA This class represents array on dppy device. Can you please elaborate on what information you want me to add? The class has comments about args and description is given in the descripton of the PR.

numba_dppy/dppy_devicearray.py
reazulhoque: Changed.

diptorupd (changes requested):
Several design questions need clarifying, refer my inline comments. My larger question is regarding how our array should use DPPY’s usm_array container once that is ready. Also, this array should support USM malloc_device.

diptorupd:
The PR description needs improvements e.g. what does “arrays in device” mean? Please use SYCL terminology as far as possible. Also, the link to numba code is not useful to a general reviewer. The exact design goals and requirements need to be spelled out.

numba_dppy/compiler.py
```diff
@@ -444,7 +446,8 @@ def _pack_argument(self, ty, val, sycl_queue, device_arr, access_type):
             # device_array or if access_type of this device_array
             # is not of type read_only and read_write
             usm_buf, usm_ndarr, orig_ndarray = device_arr
-            np.copyto(orig_ndarray, usm_ndarr)
+            if isinstance(orig_ndarray, np.ndarray):
+                np.copyto(orig_ndarray, usm_ndarr)
```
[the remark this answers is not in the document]
reazulhoque: The `DPCTLQueue_Memcpy` in parfor is using the `C` function which can work with `void *`. The `DPCTLQueue_Memcpy` we have to use here has to be the `Python` API which expects the source and destination to be of type `_Memory`. The intention here is to avoid creating a `_Memory` object wrapping the `ndarray` data buffer, I am not sure if that is possible either.

reazulhoque:
Closing this PR as dppy.kernel now supports arbitrary array with `__sycl_usm_array_interface__`.

PR #334, opened 2021-04-21
Rename dpNP to dpnp

diptorupd:


PokhodenkoSA:
@diptorupd
It is good initiative. Did you find all dpNP cases? I think DPNP should be renamed too.

PokhodenkoSA (approved):

diptorupd:
> @diptorupd
> It is good initiative. Did you find all dpNP cases? I think DPNP should be renamed too.

I think I got all the dpNP uses. Will check for DPNP.

PR #339, opened 2021-04-26
Extend itanium mangler to support numba.types.CPointer and add test

reazulhoque:
This PR will extend Numba's itanium mangler to support numba.types.CPointer with extended type qualifier address space. This work is needed for mangling native SPIRV instuctions that we need to insert at LLVM-IR level.

The extension will mangle a numba.types.CPointer as follows:
`int32 *` -> `"Pi"`
`int (address_space(1)) *` -> `"PU3AS1i"`

Please refer [here](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangling-type) for mangling rules for `pointer` type and [here](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.qualified-type) for qualifier of types.


numba_dppy/extended_numba_itanium_mangler.py
```diff
@@ -0,0 +1,66 @@
...
+    target.SPIR_GLOBAL_ADDRSPACE: "AS1",
+    target.SPIR_CONSTANT_ADDRSPACE: "AS2",
+    target.SPIR_LOCAL_ADDRSPACE: "AS3",
+    target.SPIR_GENERIC_ADDRSPACE: "AS4",
+}
+
+
+def mangle_type_or_value(typ):
+    """
+    Mangle type parameter and arbitrary value.
```
diptorupd, line 29: Thanks for adding a docstring. We need lots more of them :).

Can you change it to google style? Add `Args` and `Return`. Also add a new line after the first line to separate the short description from the detailed description.

numba_dppy/extended_numba_itanium_mangler.py
```diff
@@ -0,0 +1,66 @@
...
+    This function extends Numba provided `magle_type_or_value()` to
```
diptorupd, line 30: `Numba's` instead of `Numba provided`? Also, please make the e.g. more verbose. Such as:

For example, an ``int *`` argument will be mangled to "Pi".

numba_dppy/extended_numba_itanium_mangler.py
```diff
@@ -0,0 +1,66 @@
...
+    support numba.types.CPointer type.
+    e.g. `int *` -> "Pi"
+    For extended qualifier like `addrspace`, we generate "U" representing
```
diptorupd, line 33: We are not handling any other extended qualifier other than address spaces. I will reword it to: "Mangling of extended qualifiers is supported only for address space qualifiers. In which case, the mangling follows the rule defined in Section 5.1.5.1 of the [Itanium ABI](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.qualified-type).

numba_dppy/extended_numba_itanium_mangler.py
```diff
@@ -0,0 +1,66 @@
...
+    the presence of an extended qualifier. The actual address space is treated
+    the same way as an identifier.
+    e.g. `int (address_space(1)) *` -> "PU3AS1i"
```
diptorupd, line 36: Same changes as the previous example. Moreover, the example is neither C/C++ nor Numba. Maybe you want something like `int global *` to "PU2AS1i"?

numba_dppy/extended_numba_itanium_mangler.py
```diff
@@ -0,0 +1,66 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from numba.core import itanium_mangler, types
+from numba_dppy import target
+
+EXQ2ABI = {
```
diptorupd, line 18: Add a comment stating what the dictionary stores.

reazulhoque: I think we don't need to hardcode this values. The Itanium ABI specifies `address_space` to be mangled as `AS` and the `address_space` is assigned to the pointer so we can extract it from there.

numba_dppy/extended_numba_itanium_mangler.py
reazulhoque: Updated.

numba_dppy/extended_numba_itanium_mangler.py
reazulhoque: Updated.

numba_dppy/extended_numba_itanium_mangler.py
reazulhoque: Updated.

numba_dppy/extended_numba_itanium_mangler.py
reazulhoque: Updated.

numba_dppy/extended_numba_itanium_mangler.py
```diff
@@ -0,0 +1,66 @@
...
+def mangle_type_or_value(typ):
+    """
+    Mangle type parameter and arbitrary value.
+
+    This function extends Numba's `magle_type_or_value()` to
+    support numba.types.CPointer type.
+    e.g. an `int *` argument will be mangled to "Pi".
+    Mangling of extended qualifiers is supported only
+    for address space qualifiers. In which case, the mangling
+    follows the rule defined in Section 5.1.5.1 of the Itanium ABI.
```
diptorupd, line 28: having the link to the Itanium ABI will help.

numba_dppy/extended_numba_itanium_mangler.py
```diff
@@ -0,0 +1,66 @@
...
+from numba.core import itanium_mangler, types
+from numba_dppy import target
+
+
+def mangle_type_or_value(typ):
+    """
+    Mangle type parameter and arbitrary value.
+
+    This function extends Numba's `magle_type_or_value()` to
+    support numba.types.CPointer type.
```
diptorupd, line 24: ` This function extends Numba's ``magle_type_or_value()`` to
    support numba.types.CPointer type, e.g., an ``int *`` argument will be mangled to "Pi".`

Double back tick helps it render correctly in Sphinx (maybe we are missing some extension and that is why double back tick is needed). The other is just grammar nitpick.

diptorupd (approved):
Approved bar the two comment nitpicks.

numba_dppy/extended_numba_itanium_mangler.py
reazulhoque: Updated

PR #340, opened 2021-04-27
Feature/changes related to dpctl0.7

diptorupd:
Updates numba-dppy to use dpctl 0.7.0rc1

- Changes to `numba_dppy`:
  - `SyclQueue.equals` has been removed. Instead `==` operator now works.
  - `MemoryUSMShared` no longer allocates a buffer on the "current queue" when no queue argument is specified, it now uses the default queue as returned by DPC++'s default selector to allocate memory when no queue is specified. For this reason, the queue parameter needs to be properly set when calling the `MemoryUSMShared` constructor.
  - `dpctl.has_sycl_platforms()` has been removed. For this reason, used an alternative logic to check if a non-host SYCL device is found (same logic as the `has_sycl_platforms()` function.)
  - `dpctl.has_gpu_queues()` and `dpctl.has_cpu_queues()` have been removed. Add a new helper function to incorporate the same logic inside `numba_dppy`.
  - Behavior of `device_init.is_available()` was changed to look for a non-host SYCL device rather than an OpenCL GPU device.
  - `SyclDevice.get_device_name()` was converted to a property. All uses of the function have been updated.

- Changes to `numba_dppy.tests`:
  - dpctl now properly supports DPC++ filter strings, so the "level0" string has been changed to "level_zero" everywhere.
  - Helper functions replacing `dpctl.has_sycl_platforms()`, `dpctl.has_gpu_queues()`, `dpctl.has_cpu_queues()` were added to a new `_helper.py` module. The previous `skip_tests.py` module functions were moved to `_helper.py` and `skip_tests.py` was removed.

- Other minor changes:
  - Few docstrings were updated.

- meta.yaml changes the dpctl version requirement to `0.7.*`

Supersedes #309


reazulhoque:
>MemoryUSMShared no longer allocates a buffer on the "current queue" when no queue argument is specified, it now uses the default queue as returned by DPC++'s default selector to allocate memory when no queue is specified. For this reason, the queue parameter needs to be properly set when calling the MemoryUSMShared constructor.

Did the C-API change for this function? I ask as we create USM shared memory in case of auto offload too, do we need to take that into accout as well?

diptorupd:
> Did the behavior of the `C` function change as well? I ask as we create USM shared memory in case of auto offload too, do we need to take that into account as well?

The allocator functions in the C API need a `DPCTLSyclQueueRef` argument. So, I will expect they will work as -is. Can you point me to the location where we are calling C API memory allocator fucntions inside Numba-dppy?

reazulhoque:
You are right we are passing the queue to the `C` function. [This](https://github.com/IntelPython/numba-dppy/blob/main/numba_dppy/dppy_host_fn_call_gen.py#L131) is where we are generating a call to that function.

reazulhoque (approved):

PR #341, opened 2021-04-27
build with numpy 1.19.2 for NDA release

xaleryb:


diptorupd:
The PR is no longer relevant.

PR #353, opened 2021-05-11
Print traceback when DPPY lowering failed.

diptorupd:
Currently, when parfor lowering fails and we fall back to the default CPU lowerer for a parfor a very skimpy message is printed. The PR adds the traceback to get some more information on why and where DPPY lowering failed.


reazulhoque (commented):
The more information the better, do we want to envelope it between a debug flag?

diptorupd:
> The more information the better, do we want to envelope it inside a debug flag?

Open to suggestions. Even bare minimum information is wrapped in the debug flag.

diptorupd:
> Open to suggestions. Even bare minimum information is wrapped in the debug flag.

I added the exact device filter string based off `dpctl.get_current_queue` to the warning message that gets printed.

diptorupd:
Merging as no opposition. We can further improve the error reporting in future.

Issue #335, opened 2021-04-21
Sort all imports


diptorupd:
@PokhodenkoSA @reazulhoque @akharche @1e-to @DrTodd13 @vlad-perevezentsev 

VS Code has the nice feature to sort imports alphabetically. Is this a style that we should standardize?

PokhodenkoSA:
Sorting imports ABC is not correct. [PEP-8](https://www.python.org/dev/peps/pep-0008/#imports):
> Imports should be grouped in the following order:
> 1. Standard library imports.
> 2. Related third party imports.
> 3. Local application/library specific imports.
>
> You should put a blank line between each group of imports.

flake8 help to find correct order issues. Also [isort](https://pypi.org/project/isort/) is an auto tool for it.

diptorupd:
Can we closed as we moved to isort a while back.

PR #354, opened 2021-05-12
Fix SPIR-V validation

akharche:
This PR only affects spirv validation (NUMBA_DPPY_SPIRV_VAL=1). It resolves the problem with not appropriate debug instruction name for offloading with ocl.
If we don't add "--spirv-debug-info-version=ocl-100" flag to spirv generating, we get the following error:
`error: line 9: Invalid extended instruction import 'SPIRV.debug'`

Add OpenCL.DebugInfo.100 as flag to SPIRV generator according to specification: https://www.khronos.org/registry/spir-v/specs/unified1/OpenCL.DebugInfo.100.html

### DIFF
--- a/numba_dppy/compiler.py
+++ b/numba_dppy/compiler.py
@@ -31,12 +31,11 @@ import numpy as np
 
 from . import spirv_generator
 
-import os
 from numba.core.compiler import DefaultPassBuilder, CompilerBase
 from numba_dppy.dppy_parfor_diagnostics import ExtendedParforDiagnostics
+from numba_dppy.config import DEBUG
 
 
-DEBUG = os.environ.get("NUMBA_DPPY_DEBUG", None)
 _NUMBA_DPPY_READ_ONLY = "read_only"
 _NUMBA_DPPY_WRITE_ONLY = "write_only"
 _NUMBA_DPPY_READ_WRITE = "read_write"
--- a/numba_dppy/spirv_generator.py
+++ b/numba_dppy/spirv_generator.py
@@ -86,8 +86,9 @@ class CmdLine(object):
         # Get optimization level from NUMBA_OPT
         opt_level_option = f"-O{config.OPT}"
 
+        # spirv-debug-info-version flag is for generating right ocl debug info name
         check_call(["opt", opt_level_option, "-o", ipath + ".bc", ipath])
-        check_call(["llvm-spirv", "-o", opath, ipath + ".bc"])
+        check_call(["llvm-spirv", "--spirv-debug-info-version=ocl-100", "-o", opath, ipath + ".bc"])
 
         if dppy_config.SAVE_IR_FILES == 0:
             os.unlink(ipath + ".bc")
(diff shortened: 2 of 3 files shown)


PokhodenkoSA:
Consider to add the flag only in debug mode.

diptorupd:
@akharche please add some comments/documentation about what the flag does and why it is needed. The link to the manual is good, but a summary needs to be provided in the PR. Very hard to make out the motivation for the PR from the link to the spec.

numba_dppy/spirv_generator.py
```diff
@@ -85,9 +85,20 @@ def generate(self, ipath, opath):
 
         # Get optimization level from NUMBA_OPT
         opt_level_option = f"-O{config.OPT}"
+        if dppy_config.DEBUG:
+            llvm_spirv_command = [
+                "llvm-spirv",
+                "--spirv-debug-info-version=ocl-100",
```
PokhodenkoSA, line 91: Use variable for additional options.

numba_dppy/spirv_generator.py
```diff
@@ -85,9 +85,11 @@ def generate(self, ipath, opath):
 
         # Get optimization level from NUMBA_OPT
         opt_level_option = f"-O{config.OPT}"
+        # spirv-debug-info-version flag is for generating right ocl debug info name
+        ocl_debug_flag = ["--spirv-debug-info-version=ocl-100"] if dppy_config.DEBUG else []
```
PokhodenkoSA, line 89:
```suggestion
        llvm_spirv_flags = ["--spirv-debug-info-version=ocl-100"] if dppy_config.DEBUG else []
```

PokhodenkoSA:
```suggestion
        llvm_spirv_flags = []
        if dppy_config.DEBUG: llvm_spirv_flags.append("--spirv-debug-info-version=ocl-100")
```

numba_dppy/compiler.py
```diff
@@ -31,12 +31,11 @@
 
 from . import spirv_generator
 
-import os
 from numba.core.compiler import DefaultPassBuilder, CompilerBase
 from numba_dppy.dppy_parfor_diagnostics import ExtendedParforDiagnostics
+from numba_dppy.config import DEBUG
```
PokhodenkoSA, line 36: DEBUG variable should be removed from numba_dppy.compiler.
It will require replace callers code.
It is for separate PR.

numba_dppy/compiler.py
```diff
@@ -31,12 +31,11 @@
 
 from . import spirv_generator
 
-import os
```
PokhodenkoSA, line 34: We have a lot of places with incorrect imports. Need to apply isort and pylint in CI.

akharche:
> @akharche please add some comments/documentation about what the flag does and why it is needed. The link to the manual is good, but a summary needs to be provided in the PR. Very hard to make out the motivation for the PR from the link to the spec.

@diptorupd the PR description was updated

PokhodenkoSA:
@akharche we have no user docs about NUMBA_DPPY_SPIRV_VAL. Could you add it please?

numba_dppy/compiler.py
PokhodenkoSA: https://github.com/IntelPython/numba-dppy/issues/335

PokhodenkoSA (approved):

akharche:
@PokhodenkoSA It's a good idea to add validation to CI, but firstly we need to fix it fully

PR #355, opened 2021-05-12
Adds a new utils submodule to provide LLVM IR builder helpers.

diptorupd:
- Helper function to get the LLVM type from numba types.
  - A new helper class storing LLVM types for integers.


reazulhoque:
Are we misssing the usage of the utils?

diptorupd:
Merging as this PR is non-intrusive and no existing code is dependent on this code.

PR #362, opened 2021-05-14
Update Developer Guides about DPNP integration

PokhodenkoSA:
Status: contains all text.

To fix:
- [x] Links 
- [x] Code blocks
- [x] Add API references (requires #388)


PokhodenkoSA:
Added a lot of garbage by mistake. Will clean it.

PR #219, opened 2021-01-28
Add `import warnings` as warnings is used

PokhodenkoSA:

PR #368, opened 2021-05-18
Improve error reporting when searching for dpctl.

diptorupd:
The PR improves the user warnings that are printed when a required version of `dpctl` is not found. There are three things that are checked for `numba_dppy` to set the `dppy_found` flag:

1. `import dpctl` does not raise an error.
2. The `dpctl` version is at least 0.7.0.
3. The `dpctl.select_default_device().is_host` returns `False`, *i.e.*, there is at least one non-host SYCL device.

Closes #219.

### DIFF
--- a/numba_dppy/config.py
+++ b/numba_dppy/config.py
@@ -16,6 +16,14 @@ import os
 import warnings
 from packaging import version
 
+class DpctlMinimumVersionRequiredError(Exception):
+    """
+    A ``DpctlMinimumVersionRequiredError`` indicates that the version of dpctl
+    does not satisfy the minimum version requirement.
+
+    """
+    pass
+
 # Check for dpctl 0.7.0 or higher on the system.
 _dpctl_found = False
 try:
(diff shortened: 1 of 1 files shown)


diptorupd:
@PokhodenkoSA At least some of the `UserWarning` can be tested if we set these two environment variables: `SYCL_DEVICE_FILTER=host` and `SYCL_ENABLE_HOST_DEVICE=1`. Trying to import `numba_dppy` after setting these two environment variables will trigger the `dpctl could not find any non-host SYCL device...` warning. I can trigger it in normal Python. Do you know who to set environment variables in Pytest fixtures? I tried the following test case but that did not work:

```python
import pytest
import os


def test_import_numba_dppy_with_only_host_device():
    with pytest.warns(
        UserWarning, match=r"dpctl could not find any non-host SYCL device"
    ):
        try:
            old_environ = dict(os.environ)
            os.environ["SYCL_DEVICE_FILTER"] = "host"
            os.environ["SYCL_ENABLE_HOST_DEVICE"] = "1"
            import numba_dppy
        except ImportError:
            pass
        finally:
            os.environ.clear()
            os.environ.update(old_environ)
```

numba_dppy/config.py
```diff
@@ -13,20 +13,36 @@
...
-        # For the extension to work we should have at least one
-        # non-host SYCL device.
-        dppy_present = not dpctl.select_default_device().is_host
-    except ValueError:
-        dppy_present = False
+    # Versions of dpctl lower than 0.7.0 are not compatible with current main
+    # of numba_dppy.
+    if version.parse(dpctl.__version__) < version.parse("0.7.0"):
+        msg = "numba_dppy requires dpctl 0.7.0 or higher."
+        warnings.warn(msg, UserWarning)
```
reazulhoque, line 28: Is generating an warning the intended behavior? Will users be allowed to proceed when version is lower than 0.7.0?

reazulhoque (approved):

PR #369, opened 2021-05-18
Fixes memory leaks in the usage of dpctl C API functions.

diptorupd:
- dpctl C API's opaque pointers for sycl::event objects need
    to be properly freed. Unless freed, zeEventDestroy is never
    called and tools such as Intel Advisor fail to properly match
    call traces to be able to generate kernel profiles.
  - Similar fixes for sycl::queue objects.
  - The dpctl C API interface was refactored and moved into a new
    sub-module called `driver`.
  - Some internal function names were changed to better reflect usage.


diptorupd:
@adarshyoga Can you please test this PR out to see if Intel Advisor works as expected? It is a cleaned up version of the rough patch I had given you a while back.

reazulhoque (approved):
Much needed refactoring work, thank you for taking it on. Ideally we would want to separate the PR to one containing the refactoring and one containing the leak. Few pending questions:

1. Can you please point where we have added the code that fixes the leak?
2. Is there a way to check that we are not suffering from this kinds of leaks?

adarshyoga (approved):

PR #370, opened 2021-05-18
Refactoring spirv_generator.py

PokhodenkoSA:
Code simplification.

### DIFF
--- a/numba_dppy/spirv_generator.py
+++ b/numba_dppy/spirv_generator.py
@@ -42,40 +42,28 @@ def check_call(*args, **kwargs):
 
 class CmdLine(object):
     def disassemble(self, ipath, opath):
-        check_call(
-            [
-                "spirv-dis",
-                # "--no-indent",
-                # "--no-header",
-                # "--raw-id",
-                # "--offsets",
-                "-o",
-                opath,
-                ipath,
-            ]
-        )
+        flags = []
+        # flags.append("--no-indent")
+        # flags.append("--no-header")
+        # flags.append("--raw-id")
+        # flags.append("--offsets")
+        check_call(["spirv-dis", *flags, "-o", opath, ipath])
 
     def validate(self, ipath):
         check_call(["spirv-val", ipath])
 
     def optimize(self, ipath, opath):
-        check_call(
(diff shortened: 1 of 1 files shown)


numba_dppy/spirv_generator.py
```diff
@@ -42,61 +42,49 @@ def check_call(*args, **kwargs):
...
-                # "--no-header",
-                # "--raw-id",
-                # "--offsets",
-                "-o",
-                opath,
-                ipath,
-            ]
-        )
+        flags = []
+        # flags.append("--no-indent")
```
diptorupd, line 46: Let us get rid of the commented out lines.

numba_dppy/spirv_generator.py
```diff
@@ -42,61 +42,49 @@ def check_call(*args, **kwargs):
...
-                # "--inline-entry-points-exhaustive",
-                # "--flatten-decorations",
-                # "--compact-ids",
-                "-o",
-                opath,
-                ipath,
-            ]
-        )
+        flags = []
+        # flags.append("--strip-debug")
```
diptorupd, line 57: same here.

numba_dppy/spirv_generator.py
```diff
@@ -42,61 +42,49 @@ def check_call(*args, **kwargs):
...
 
+        ipath_optimized = ipath + ".bc"
+
         # Get optimization level from NUMBA_OPT
-        opt_level_option = f"-O{config.OPT}"
+        opt_flags = [f"-O{config.OPT}"]
 
-        check_call(["opt", opt_level_option, "-o", ipath + ".bc", ipath])
-        check_call(["llvm-spirv", "-o", opath, ipath + ".bc"])
+        check_call(["opt", *opt_flags, "-o", ipath_optimized, ipath])
```
diptorupd, line 79: Ideally we should use `llvm-as` instead of `opt`.

numba_dppy/spirv_generator.py
PokhodenkoSA: So I need another way to provide info for which flags are possible.

PR #373, opened 2021-05-19
Add assertion to make sure data and kernel are using the same SYCL context

reazulhoque:
This PR ensures USM data and SYCL kernels are using the same SYCL context.

### DIFF
--- a/numba_dppy/compiler.py
+++ b/numba_dppy/compiler.py
@@ -405,8 +405,8 @@ class DPPYKernel(DPPYKernelBase):
 
         kernelargs = []
         internal_device_arrs = []
-        for ty, val, access_type in zip(
-            self.argument_types, args, self.ordered_arg_access_types
+        for idx, (ty, val, access_type) in enumerate(
+            zip(self.argument_types, args, self.ordered_arg_access_types)
         ):
             self._unpack_argument(
                 ty,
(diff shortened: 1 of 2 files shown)


numba_dppy/compiler.py
```diff
@@ -493,6 +494,12 @@ def _unpack_argument(
 
         if isinstance(ty, types.Array):
             if hasattr(val.base, "__sycl_usm_array_interface__"):
+                assert (
+                    sycl_queue.get_sycl_context() == val.base._queue.get_sycl_context()
```
diptorupd, line 498: Couple of issues:

a) The behavior is going to change when we move to the compute-follows-data design. In the current setup, I think the check is fine, but we can also do a copy through host if we want.
b) dpctl 0.8.0 introduced a new property `sycl_context` in the `MemoryUSMShared` class specifically for this scenario.

We can either merge the change and I can update it along with any other change for dpctl 0.8.0, or we hold off mering till the dpctl 0.8.0 chnages PR is ready.

reazulhoque: I would like to propose that we merge it now. Once 0.8 is ready the API can be easily updated as APIs will need to be updated throughout Numba_dppy.

PokhodenkoSA:
@diptorupd @reazulhoque 
Is it actual yet?

PokhodenkoSA (changes requested):
Need to be updated.

reazulhoque:
dppy.kernel has CFD implemented. We can get rid of these checks.

PR #376, opened 2021-05-20
Workaround for unsafe address space cast

akharche:
This pr enables switching to O0 opt level for debug purposes.


akharche:
Fixed by https://github.com/IntelPython/numba-dppy/pull/432

PR #389, opened 2021-05-24
Add note about Intel Python Numba

PokhodenkoSA:


docs/user_guides/getting_started.rst
```diff
@@ -16,6 +16,10 @@ Numba-dppy depends on following components:
 * `scipy`_ (for testing)
 * `pytest`_ (for testing)
 
+.. note::
+    Numba-dppy will not work with stock `Numba`_. It requires `Intel Python Numba`_.
```
diptorupd, line 20: I will reword it to: "Numba-dppy does not yet work with stock `Numba`_. For now, we need a patched Numba version included in `Intel Python Numba`_."

We have the support in works, let people not think we will never support stock Numba :)

diptorupd (approved):
LGTM with the minor edit.

PR #390, opened 2021-05-24
Update SYCL Filter String

reazulhoque:
This PR updates the filter string in `conftest.py`


diptorupd (approved):
Thanks!

PokhodenkoSA:
@reazulhoque please provide reference to list of all available filter string (from DPC++ docs or other docs). Without it these strings are magic. How to sure that you are correct now?

PR #393, opened 2021-05-25
Fix/dppy rt extension

diptorupd:
The `dppy_rt.c` extension module stopped working with dpctl 0.8.0rc1 due to the introduction of soversion names in dpctl's `libDPCTLSyclInterface`. The PR fixes the problem by properly adding the libDPCTLSyclInterface to the rpath using setup.py. Additionally, the extension itself was refactored to not use `dlopen` but rather properly include dpctl C API headers.

The refactored extension is renamed and moved into the driver sub-module.


reazulhoque (approved):

PokhodenkoSA:
> The PR fixes the problem by properly adding the libDPCTLSyclInterface to the rpath using setup.py.

I think conda build process removes rpath. It could be the problem.

PR #396, opened 2021-05-25
Docs how set up machine for gdb

1e-to:


docs/user_guides/set_up_machine.md
```diff
@@ -0,0 +1,59 @@
...
+
+2) You can also install NEO drivers locally. In that folder with NEO, run the following commands: 
+
+```bash
+for file in `ls *.deb`; do dpkg -x $file .; done
+export MY_ACTIVE_NEO=/path/to/my/neo/usr/local/lib
+export LD_LIBRARY_PATH=${MY_ACTIVE_NEO}:${MY_ACTIVE_NEO}/intel-opencl:$LD_LIBRARY_PATH
+```
+
+You will also need to add environment variables to change the behavior of the ICD. Info about [OCL_ICD_...](https://github.com/KhronosGroup/OpenCL-ICD-Loader).
```
akharche, line 32: What is ICD and why I need to change the behavior? It's not clear

docs/user_guides/set_up_machine.md
```diff
@@ -0,0 +1,59 @@
...
+sudo modinfo igfxdcd
+```
+
+Also, you must remove the driver from the system if you want to install a different version:
+
+```bash
+sudo dpkg -r igfxdcd
+```
+
+If you are installing dsd for the first time, you need to create keys. More details [here](https://software.intel.com/content/www/us/en/develop/documentation/get-started-with-debugging-dpcpp-linux/top.html).
```
akharche, line 59: dcd??

akharche: I think, here it's enough to give a link to official docs

PR #404, opened 2021-05-26
[WIP] Accept dpctl.tensor.usm_ndarray as a valid type for @dppy.kernel

reazulhoque:
We add support for `dpctl.tensor.usm_ndarray` in `@dppy.kernel`. Users can now allocate a `usm_ndarray` and pass it to `@dppy.kernel`. Example is added in `numba_dppy/examples/usm_ndarray.py`


reazulhoque:
Support for arbitrary array in `@numba_dppy.kernel` with `__sycl_usm_array_interface__` attribute is already implemented. We do not need multiple array containers now.

PR #412, opened 2021-05-31
Fix regression in spirv generator

akharche:
Flag for debug was removed in previous prs

### DIFF
--- a/numba_dppy/spirv_generator.py
+++ b/numba_dppy/spirv_generator.py
@@ -105,7 +105,7 @@ class CmdLine(object):
             llvm_spirv_call_args = ["llvm-spirv"]
         if llvm_spirv_args is not None:
             llvm_spirv_call_args += llvm_spirv_args
-        llvm_spirv_call_args += ["-o", opath, ipath + ".bc"]
+        llvm_spirv_call_args += [*llvm_spirv_flags, "-o", opath, ipath + ".bc"]
         check_call(llvm_spirv_call_args)
 
         if dppy_config.SAVE_IR_FILES == 0:


PokhodenkoSA:
Code around is hard to read. Need refactoring.

PokhodenkoSA:
Is it possible to make test for it?

numba_dppy/tests/test_debug.py
```diff
@@ -70,3 +71,39 @@ def foo(x):
...
 
         assert expect == got
+
+
+def test_function_with_debuginfo(offload_device):
+    """
+    Check dppy function call if debug parameter is set to True
+    """
+
+    dppy.config.DEBUG = 1  # Run test in debug mode
```
PokhodenkoSA, line 81: Where is return to 0?

numba_dppy/spirv_generator.py
```diff
@@ -105,7 +105,7 @@ def generate(self, ipath, opath, llvm_spirv_args):
             llvm_spirv_call_args = ["llvm-spirv"]
         if llvm_spirv_args is not None:
             llvm_spirv_call_args += llvm_spirv_args
-        llvm_spirv_call_args += ["-o", opath, ipath + ".bc"]
+        llvm_spirv_call_args += [*llvm_spirv_flags, "-o", opath, ipath + ".bc"]
```
PokhodenkoSA, line 108: I have reverted this change and test PASSED. Why it could be?

PokhodenkoSA (changes requested):
Test does not test changes.

numba_dppy/spirv_generator.py
akharche: Because (as I said earlier) this flag only fixes error in spirv-val mode.

PokhodenkoSA: > spirv-val mode

Could you enable this mode in test?

akharche:
> The test does not test changes.

Yes. It tests calling function with debug in general. Currently it is hardly possible to write a test on checking failing in spirv-val mode because there are lots of other problems.

numba_dppy/spirv_generator.py
akharche: Yes, I can, but checking spirv (spirv-val mode) is blocked by a set of problems. This flag fixes only one problem connected to appropriate Debug name in spirv

PokhodenkoSA:
@akharche what is the status now?

PR #414, opened 2021-05-31
Implement NumPy functions `diag` via DPNP

Rubtsowa:


PokhodenkoSA (approved):

PokhodenkoSA (changes requested):
Does not work now. Requires debugging.

numba_dppy/dpnp_glue/dpnp_array_creations_impl.py
```diff
@@ -226,6 +225,82 @@ def dpnp_impl(a, b):
...
+        if v.ndim == 1:
+            n = v.shape[0] + abs(k)
+
+            shape_result = (n, n)
+        else:
+            n = min(v.shape[0], v.shape[0] + k, v.shape[1], v.shape[1] - k)
+            if n < 0:
+                n = 0
+
+            shape_result = (n,)
```
chudur-budur, line 274: Why don't we just do like
```python
if v.ndim < 1:
    raise ValueError("ndim has to be nonempty, i.e. >= 1")
elif v.ndim == 1:
    # do for ndim == 1
else:
    # do for ndim > 1  
```

numba_dppy/dpnp_glue/dpnp_array_creations_impl.py
```diff
@@ -226,6 +225,82 @@ def dpnp_impl(a, b):
...
+            raise ValueError("Not supported")
+
+        if v.ndim == 1:
+            n = v.shape[0] + abs(k)
+
+            shape_result = (n, n)
+        else:
+            n = min(v.shape[0], v.shape[0] + k, v.shape[1], v.shape[1] - k)
+            if n < 0:
+                n = 0
```
chudur-budur, line 272: If `k > 0`, `v.shape[0] - k` will be the minimum. Again if `k < 0`, `v.shape[0] + k` will be the minimum. Therefore, why don't we do like:
```python
n = v.shape[0] - k if k > 0 else v.shape[0] + k  
```
or we can just do simply like
```python
p,q = v.shape[0] - k, v.shape[0] + k
n = (p if p > 0 else 0) if k > 0 else (q if q < 0 else 0)
```

chudur-budur (commented):
Looks good, but some improvements can be done.

If you could "Allow edits from maintainers", we can rebase on the current main and test it. 
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

Also, it would be great if you could squash multiple commits into one.
https://www.scraggo.com/how-to-squash-commits/

numba_dppy/dpnp_glue/dpnp_array_creations_impl.py
chudur-budur: Also if `k > 0`, `v.shape[0] - k` will be the minimum. Again if `k < 0`, `v.shape[0] + k` will be the minimum. Therefore, why don't we do like:
```python
if k > 0:
    n = v.shape[0] if v.shape[0] <= v.shape[1] - k else v.shape[1] - k
else:
    n = v.shape[1] if v.shape[1] <= v.shape[0] + k else v.shape[0] + k  
```

diptorupd:
The code is too out of date by now.

PR #424, opened 2021-06-07
Fix debuginfo

### LINKED ISSUE #352: DEBUG: Step inside function calls (command “step”)
We can not debug nested functions.
Check where debug info missed. 
Investigate how to receive debug info on each step of pipeline.

Collected info about steps will help to write docs about compilation pipeline #385.

- [x] implement and test #424 
- [x] update docs

1e-to:
Provides the ability to generate debug information for nested functions

We need to support 2 cases:
- [x] `@kernel(debug=True)` - via decorator, it means propagate flag via function arguments or target context
- [x] `NUMBA_DPPY_DEBUGINFO=1` - via env variable, it means in some point use env variable value if the value is not provided in decorator


numba_dppy/compiler.py
```diff
@@ -196,6 +197,8 @@ def compile_kernel_parfor(sycl_queue, func_ir, args, args_with_addrspaces, debug
 
 
 def compile_dppy_func(pyfunc, return_type, args, debug=False):
+    if DEBUG:
+        debug = True
```
AlexanderKalistratov, line 201: That doesn't feels right. Why we are not passing correct `debug` flag instead?

AlexanderKalistratov (changes requested):

numba_dppy/compiler.py
akharche: Let's remove DEBUG from compiler.py and move to the place, where compile_dppy_func is called?

numba_dppy/compiler.py
```diff
@@ -166,6 +166,7 @@ def compile_kernel(sycl_queue, pyfunc, args, access_types, debug=False):
 
 def compile_kernel_parfor(sycl_queue, func_ir, args, args_with_addrspaces, debug=False):
     if DEBUG:
+        debug = True
```
PokhodenkoSA, line 169: Consider `NUMBA_DEBUG` and `NUMBA_DEBUGINFO`

numba_dppy/compiler.py
```diff
@@ -98,7 +98,7 @@ def compile_with_dppy(pyfunc, return_type, args, debug):
 
     flags = compiler.Flags()
     # Do not compile (generate native code), just lower (to LLVM)
-    if debug:
+    if DEBUGINFO:
```
AlexanderKalistratov, line 101: we already have `debug` flag passed to the function. Why are we ignoring it?
Why we are not passing correct flag?

numba_dppy/config.py
```diff
@@ -97,4 +98,5 @@ def _readenv(): ...
 NATIVE_FP_ATOMICS = _readenv("NUMBA_DPPY_ACTIVATE_ATOMCIS_FP_NATIVE", int, 0)
 LLVM_SPIRV_ROOT = _readenv("NUMBA_DPPY_LLVM_SPIRV_ROOT", str, "")
 # Emit debug info
-DEBUG = os.environ.get("NUMBA_DPPY_DEBUG", None)
+DEBUG = _readenv("NUMBA_DPPY_DEBUG", int, 0)
+DEBUGINFO = _readenv("NUMBA_DPPY_DEBUGINFO", int, DEBUGINFO_DEFAULT)
```
AlexanderKalistratov, line 102: Why do we need special environment variable for dppy? Why not use numba default?

numba_dppy/compiler.py
1e-to: This is a temporary solution, because in many functions that call compile_with_dppy no debug is specified at all, as a result of which debug=False
Further study is needed why debug is not indicated in some places and where it should come from

AlexanderKalistratov: I don't think we can accept current implementation

1e-to: @reazulhoque , Can you please tell why "debug" was added to functions parameter and for what purpose?

numba_dppy/tests/test_debug_func.py
```diff
@@ -0,0 +1,126 @@
+#! /usr/bin/env python
```
akharche, line 1: Why do we need to create a separate file? we already have "test_debug.py"

1e-to:
Fixed, @akharche , @PokhodenkoSA , @AlexanderKalistratov  please check

numba_dppy/config.py
PokhodenkoSA, line 102: Why `DEBUGINFO` when `DEBUGINFO_DEFAULT` in numba?

numba_dppy/config.py
```diff
@@ -97,4 +98,5 @@ def _readenv(): ...
 NATIVE_FP_ATOMICS = _readenv("NUMBA_DPPY_ACTIVATE_ATOMCIS_FP_NATIVE", int, 0)
 LLVM_SPIRV_ROOT = _readenv("NUMBA_DPPY_LLVM_SPIRV_ROOT", str, "")
 # Emit debug info
-DEBUG = os.environ.get("NUMBA_DPPY_DEBUG", None)
+DEBUG = _readenv("NUMBA_DPPY_DEBUG", int, 0)
```
PokhodenkoSA, line 101: Reuse default value from numba.

numba_dppy/tests/test_debuginfo.py
```diff
@@ -142,3 +142,83 @@ def foo(arr):
...
+        result = a + b
+        return result
+
+    @dppy.kernel(debug=True)
+    def data_parallel_sum(a, b, c):
+        i = dppy.get_global_id(0)
+        c[i] = func_sum(a[i], b[i])
+
+    ir_tag_func_sum = r'\!DISubprogram\(name: "test_debug_flag_generates_ir_with_debuginfo_for_func.<locals>.func_sum"'
+    ir_tag_data_parallel_sum = r'\!DISubprogram\(name: "test_debug_flag_generates_ir_with_debuginfo_for_func.<locals>.data_parallel_sum"'
```
PokhodenkoSA, line 169: Regular expression depends on test name. Imagine somebody renames the test and the test fails.

numba_dppy/tests/test_debuginfo.py
```diff
@@ -142,3 +142,83 @@ def foo(arr):
...
+        result = a + b
+        return result
+
+    @dppy.kernel
+    def data_parallel_sum(a, b, c):
+        i = dppy.get_global_id(0)
+        c[i] = func_sum(a[i], b[i])
+
+    ir_tag_func_sum = r'\!DISubprogram\(name: "test_env_var_generates_ir_with_debuginfo_for_func.<locals>.func_sum"'
+    ir_tag_data_parallel_sum = r'\!DISubprogram\(name: "test_env_var_generates_ir_with_debuginfo_for_func.<locals>.data_parallel_sum"'
```
PokhodenkoSA, line 207: The same as in comment above about depending on test name.

numba_dppy/tests/test_debuginfo.py
```diff
@@ -142,3 +142,83 @@ def foo(arr):
...
+    ir_tags = (ir_tag_func_sum, ir_tag_data_parallel_sum)
+
+    dppy.compiler.DEBUGINFO = 1
```
PokhodenkoSA, line 210: All config variables should be accessed via `config` module.

numba_dppy/tests/test_debuginfo.py
1e-to: We have no way to test correctly without the full name

numba_dppy/tests/test_debuginfo.py
1e-to: If we change the variables from the config module it won't work. Because they are read 1 time before we import numba-dppy

numba_dppy/tests/test_debuginfo.py
PokhodenkoSA: Why? This is regular expression. Just use `.*` instead of test name.

numba_dppy/tests/test_debuginfo.py
PokhodenkoSA: Please, explain in details. I think it is problem in imports (create ticket in this way) and we do not have to do workaround.

numba_dppy/compiler.py
```diff
@@ -98,7 +98,7 @@ def compile_with_dppy(pyfunc, return_type, args, debug):
 
     flags = compiler.Flags()
     # Do not compile (generate native code), just lower (to LLVM)
-    if debug:
+    if debug or DEBUGINFO_DEFAULT:
```
AlexanderKalistratov, line 101: I don't like that in the middle of pipeline we access some global variable which modifies  behavior.
Much more clean solution would be **to pass correct debug flag to the function**.
On the other hand we have at least 3 decorators, and to pass correct flag we need to modify all of them, but it all comes to this point.
On third hand, our behavior of ```DEGUGINFO``` environment variable is different from Numba behavior. In Numba ```debug``` flag overrides ```DEBUGINFO```. In our case it's vice versa.
@PokhodenkoSA what do you think?

numba_dppy/tests/test_debuginfo.py
```diff
@@ -142,3 +142,83 @@ def foo(arr):
...
+
+    with dpctl.device_context(offload_device) as sycl_queue:
+        sig = (
+            types.float32[:],
+            types.float32[:],
+            types.float32[:],
+        )
+        kernel_ir = get_kernel_ir(sycl_queue, data_parallel_sum, sig, debug=True)
+
+    dppy.compiler.DEBUGINFO_DEFAULT = 0
```
AlexanderKalistratov, line 220: We should return previous state of ```dppy.compiler.DEBUGINFO_DEFAULT```, not just set it to ```0```

Also we are not testing environment variable, only ```dppy.compiler.DEBUGINFO_DEFAULT``` flag. But I think it is ok.
@PokhodenkoSA what do you think?

numba_dppy/decorators.py
```diff
@@ -67,24 +67,31 @@ def _wrapped(pyfunc):
...
     def _wrapped(pyfunc):
-        return compile_dppy_func(pyfunc, restype, argtypes)
+        return compile_dppy_func(pyfunc, restype, argtypes, debug=debug)
 
     return _wrapped
 
 
-def _func_autojit(pyfunc):
-    return compile_dppy_func_template(pyfunc)
+def _func_autojit_wrapper(debug=False):
```
AlexanderKalistratov, line 89: I'd move this wrapper to ```func``` decorator. But it is also fine as is.

numba_dppy/tests/test_debuginfo.py
1e-to: Created https://github.com/IntelPython/numba-dppy/issues/442

numba_dppy/tests/test_debuginfo.py
PokhodenkoSA: Return to previous value +1.
Test via config value +1. Not necessary to test how env variable is transformed to config value.

numba_dppy/compiler.py
PokhodenkoSA: Yes, the behavior which is different from numba will confuse users.
Search in numba:
1. There is cool context manager `override_config` for testing config altering, we can create the same for our config
https://github.com/IntelPython/numba/blob/c4f8579f620a38e0d6d6c2e4f5b528899cdcefca/numba/tests/support.py#L515-L515
2. I find 2 places for using `DEBUGINFO_DEFAULT`
https://github.com/IntelPython/numba/blob/0c33c333443971b6d124e3f982238392423a28fb/numba/core/cpu.py#L288-L288
https://github.com/IntelPython/numba/blob/dd57c669e2b8c2dfcdbd2952ea99fe6494b71874/numba/np/ufunc/ufuncbuilder.py#L41-L41
Both are applied for creating default options. In our case it is `compiler.Flags()`. We can inherit from it.
The point is that default `debug` argument value should be `None`. And if it is not None then flag `debuginfo` in `compiler.Flags()` should be forced to it.

PokhodenkoSA: So `debug` argument should be `None` (i.e. undefined) by default.

numba_dppy/compiler.py
```diff
@@ -212,11 +212,11 @@ class dppy_function_template(ConcreteTemplate):
 
 
 # Compile dppy function template
-def compile_dppy_func_template(pyfunc):
+def compile_dppy_func_template(pyfunc, debug=False):
```
PokhodenkoSA, line 215:
```suggestion
def compile_dppy_func_template(pyfunc, debug=None):
```

numba_dppy/decorators.py
```diff
@@ -67,24 +67,31 @@ def _wrapped(pyfunc):
     return _wrapped
 
 
-def func(signature=None):
+def func(signature=None, debug=False):
```
PokhodenkoSA, line 70:
```suggestion
def func(signature=None, debug=None):
```

numba_dppy/compiler.py
AlexanderKalistratov: Setting `debuginfo` to default value inside `compiler.Flags()` sounds like a great idea, if possible

numba_dppy/decorators.py
```diff
@@ -67,24 +67,31 @@ def _wrapped(pyfunc):
...
         func = signature
-        return _func_autojit(func)
+        return _func_autojit(func, debug=debug)
     else:
-        return _func_jit(signature)
+        return _func_jit(signature, debug=debug)
 
 
-def _func_jit(signature):
+def _func_jit(signature, debug=False):
```
PokhodenkoSA, line 80:
```suggestion
def _func_jit(signature, debug=None):
```

numba_dppy/decorators.py
PokhodenkoSA, line 89:
```suggestion
def _func_autojit_wrapper(debug=None):
```

numba_dppy/decorators.py
```diff
@@ -67,24 +67,31 @@ def _wrapped(pyfunc):
...
-def _func_autojit(pyfunc):
-    return compile_dppy_func_template(pyfunc)
+def _func_autojit_wrapper(debug=False):
+    def _func_autojit(pyfunc, debug=debug):
+        return compile_dppy_func_template(pyfunc, debug=debug)
+
+    return _func_autojit
+
+
+def _func_autojit(pyfunc, debug=False):
```
PokhodenkoSA, line 96:
```suggestion
def _func_autojit(pyfunc, debug=None):
```

PokhodenkoSA (changes requested):

numba_dppy/tests/test_debuginfo.py
PokhodenkoSA: #424 does not answer my question about using `config` module instead of using `compiler` module.

1e-to: In the config module, the behavior is the same as with environment variables, and the variable from the compiler can be changed after importing dppy

AlexanderKalistratov: @1e-to so we have TWO different `DEBUGINFO` flags? One in `config` and one in `compiler`?

AlexanderKalistratov: It doesn't looks so...

numba_dppy/compiler.py
PokhodenkoSA: > Setting debuginfo to default value inside compiler.Flags()

In separate PR.

numba_dppy/tests/test_debuginfo.py
PokhodenkoSA: Fixed.

PokhodenkoSA (approved):
LGTM

PR #428, opened 2021-06-08
Use default device in examples

### LINKED ISSUE #427: Use default device context in examples
All examples contain code for selecting device context. It is not how we recommend to code.
The better approach is to use default device context and provide possibility to select default via environment variable.

See https://github.com/IntelPython/numba-dppy/pull/426/files#r647426318.

PokhodenkoSA:
Fix #427 

- [x] Use `dpctl.select_default_device()` in all examples. Use `SYCL_DEVICE_FILTER` env var outside of examples.
- [x] Add print `Done...` for each example
- [x] Remove `try/except` as if SYCL device unavailable, then error or exception should be nice.

@diptorupd @reazulhoque @1e-to 
If you like it, someone can extend this to all examples.


reazulhoque:
@PokhodenkoSA Can we close this as #430 supersedes this?

PokhodenkoSA:
> Can we close this as #430 supersedes this?

As I understand the approach is accepted. I will do update in #430.

PokhodenkoSA:
Reopened for collecting only device contxt changes. @reazulhoque It is ready for review.

PokhodenkoSA:
Ready for review.

scripts/run_examples.sh
```diff
@@ -0,0 +1,49 @@
...
+  check numba_dppy/examples/debug/dppy_func.py
+  check numba_dppy/examples/debug/sum.py
+}
+
+run_with_device() {
+  echo "Run with SYCL_DEVICE_FILTER=$1 ..."
+  SYCL_DEVICE_FILTER=$1 run_checks
+}
+
+run_with_device level_zero:gpu:0
```
1e-to, line 45: Do we only want to test on these three configurations?

PokhodenkoSA: I want to modify this script for running in CI (in separate PR). In this case `SYCL_DEVICE_FILTER` will be used from outside, because we do not know exact environment in which script will be run.

scripts/run_examples.sh
```diff
@@ -0,0 +1,49 @@
...
+  check numba_dppy/examples/auto_offload_examples/sum-4d.py
+  check numba_dppy/examples/auto_offload_examples/sum-5d.py
+
+  check numba_dppy/examples/debug/dppy_func.py
+  check numba_dppy/examples/debug/sum.py
+}
+
+run_with_device() {
+  echo "Run with SYCL_DEVICE_FILTER=$1 ..."
+  SYCL_DEVICE_FILTER=$1 run_checks
```
PokhodenkoSA, line 42: SYCL_DEVICE_FILTER - filters, not actually change default device.

PokhodenkoSA: Created separate ticket for this #438.

PokhodenkoSA: Related https://github.com/IntelPython/numba-dppy/pull/430#issuecomment-858072151

PR #430, opened 2021-06-08
Proposals for improving examples

### LINKED ISSUE #427: Use default device context in examples

PokhodenkoSA:
- Closes #428 - default device. Fixes #427.
- no `random`, use `arange`
- no `ones_like`, use `empty_like`

### DIFF
--- a/numba_dppy/examples/sum.py
+++ b/numba_dppy/examples/sum.py
@@ -42,9 +42,9 @@ def main():
     N = global_size
     print("N", N)
 
-    a = np.array(np.random.random(N), dtype=np.float32)
-    b = np.array(np.random.random(N), dtype=np.float32)
-    c = np.ones_like(a)
+    a = np.arange(N, dtype=np.float32)
+    b = np.arange(N, dtype=np.float32)
+    c = np.empty_like(a)
 
     # Use the environment variable SYCL_DEVICE_FILTER to change the default device.
     # See https://github.com/intel/llvm/blob/sycl/sycl/doc/EnvironmentVariables.md#sycl_device_filter.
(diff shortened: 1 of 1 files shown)


reazulhoque:
@PokhodenkoSA I like the overall direction, can you please add the same default device selector in all examples in the same PR?

PokhodenkoSA:
> can you please add the same default device selector in all examples in the same PR?

Yes. I will do it.

diptorupd:
In general, using `select_default_device` is fine as far as making the examples shorter. However, I want to be careful as to advertising the use of `SYCL_DEVICE_FILTER` environment variable. Depending on how the variable is set up it may not only set up a default device, but also disable devices. E.g., `export SYCL_DEVICE_FILTER="level_zero:gpu"` will make all OpenCL device non-discoverable. For this reason, I do not want users to think that is the way we want them to control devices. 

Instead, we should show in some example that if a user wants to use something other than the default device, they should use the appropriate device selector from dpctl. Moreover, at least some examples should cover the use case where we show the same code running on multiple type of devices by changing the `device_context`.

PokhodenkoSA:
> at least some examples should cover the use case where we show the same code running on multiple type of devices by changing the `device_context`.

I will create separate example for this.

diptorupd (approved):

PokhodenkoSA:
Do not merge until I apply this changes to all examples.

diptorupd:
The changes suggested in the PR are already implemented in the main branch,

PR #432, opened 2021-06-09
Fix/remove addressspace cast

diptorupd:
The PR introduces changes to remove the address space casting from `Global` to `Generic` in the wrapper function for kernels.

- Cast `numba.core.npytypes.Array` to `DppyArray` type. The `DppyArray` type defines the `meminfo` and `parent` attributes as `CPointer` types rather than `Pyobject` type, allowing us to specifying an address space for the pointers.
- Change the default address space for `CPointer` to `Global` from `Generic` and remove the casting of `Global` pointers to `Generic` pointers.


diptorupd:
@akharche Please try this branch. I need to clean up lots of things (commented code, undocumented code, *etc.*) but see if the approach works for us.

akharche:
@diptorupd Checked this fix, **it solves the problem** with O0 running. Also, I can see all the variables names when run GDB.

PokhodenkoSA:
#421 merged to main. It contains test for this PR with XPASS. I propose to merge main to this PR.

PokhodenkoSA (commented):
Solution works. I think `target.py` is not clean yet. But I like the rest.

PokhodenkoSA:
@akharche 
I have tested locally `test_debug_info_locals_vars_on_no_opt` with different `OPT` values and it seems that debug info for local vars present with both `OPT=0` and `OPT=3`.

PokhodenkoSA (approved):

PR #435, opened 2021-06-09
Added local addrspace testing for native atomic op

reazulhoque:
This PR extends the native atomic ops on local addrspace variables.


PokhodenkoSA (approved):
LGTM

PR #441, opened 2021-06-11
Tests improvements for debuginfo

PokhodenkoSA:
Proposed improvements:
1. use `assert tag in ir` - it shows good explanation, but requires `tag` to be simple string
2. if RE not necessary then use simple strings
3. get `sycl_queue` from `dpctl.get_current_queue()` - device context is not necessary
4. test on default device - IR does not depend on device - 1 test instead of 3 tests

### DIFF
--- a/numba_dppy/tests/test_debuginfo.py
+++ b/numba_dppy/tests/test_debuginfo.py
@@ -71,7 +71,7 @@ def test_debug_flag_generates_ir_with_debuginfo(debug_option):
         assert tag not in kernel_ir
 
 
-def test_debug_info_locals_vars_on_no_opt(offload_device):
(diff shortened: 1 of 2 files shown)


PokhodenkoSA:
@diptorupd could you check my assumptions 3 and 4 in description?

1e-to (approved):

PokhodenkoSA:
As it is accepted, I will apply it to all cases in tests.

akharche (approved):

diptorupd:
> 
> 
> @diptorupd could you check my assumptions 3 and 4 in description?

Yes, you are correct. The changes make sense in the present test case.

PR #443, opened 2021-06-11
Fix env variables in debug tests

akharche:
Fix env variables in debug tests

### DIFF
--- a/numba_dppy/tests/test_debuginfo.py
+++ b/numba_dppy/tests/test_debuginfo.py
@@ -15,6 +15,7 @@
 
 import re
 import pytest
+import os
 
 import dpctl
 import numpy as np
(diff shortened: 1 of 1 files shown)


PokhodenkoSA:
I have implemented the same in #432 via `override_config`.

numba_dppy/tests/test_debuginfo.py
```diff
@@ -99,6 +100,7 @@ def foo(var_a, var_b, var_c):
 
     ir_tags = (ir_tag_var_a, ir_tag_var_b, ir_tag_var_c, ir_tag_var_i)
 
+    opt_curr_value = os.environ.get("NUMBA_OPT", 3)  # 3 is a default value
```
PokhodenkoSA, line 103: Do not use both `os.environ` and `config`. They do the same from different levels.

PokhodenkoSA (changes requested):

akharche:
Resolved by #432

PR #452, opened 2021-06-15
Update docs about debugger local vars

akharche:
Update docs about debugger local vars

### DIFF
--- a/docs/user_guides/debugging/local_variables.rst
+++ b/docs/user_guides/debugging/local_variables.rst
@@ -6,32 +6,11 @@ Local variables
     - :samp:`NUMBA_OPT=0` "no optimization" level - all local variables of the kernel function are available.
     - :samp:`NUMBA_OPT=1` or higher - some variables may be optimized out.
 
-Consider `numba-dppy` kernel code from `sum.py` example:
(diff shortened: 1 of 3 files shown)


docs/user_guides/debugging/local_variables.rst
```diff
@@ -4,89 +4,94 @@ Local variables
...
+
+    (gdb) print i
+    $2 = 1
+
 
 .. note::
 
     Known issues:
-      - Kernel variables are shown in IR representation.
+      - Kernel variables names are shown in IR representation.
```
PokhodenkoSA, line 97:
```suggestion
      - Kernel variable names are shown in IR representation.
```

PokhodenkoSA (changes requested):
Please, include example code from file in numba_dppy/examples

docs/user_guides/debugging/local_variables.rst
mpuputti: Is it clear to a reader, what is "IR representation"? If not providing an explanation in this page, perhaps we could add a link to a page which explains / shows it to the user?

PokhodenkoSA (approved):

PR #456, opened 2021-06-16
Add docs about backtrace

1e-to:
Add docs about backtrace

### DIFF
--- a/docs/user_guides/debugging/backtrace.rst
+++ b/docs/user_guides/debugging/backtrace.rst
@@ -1,0 +1,44 @@
+Backtrace
+==========
+
+``backtrace``
+-------------
+
+Let's consider the work of the command ``backtrace`` in the following example ``numba_dppy/examples/debug/simple_dppy_func.py``:
+
+.. literalinclude:: ../../../numba_dppy/examples/debug/simple_dppy_func.py
+    :lines: 15-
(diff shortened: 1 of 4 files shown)


docs/user_guides/debugging/backtrace.rst
```diff
@@ -0,0 +1,44 @@
...
+.. literalinclude:: ../../../numba_dppy/examples/debug/simple_dppy_func.py
+    :lines: 15-
+    :linenos:
+
+Run debugger:
+
+.. code-block:: bash
+
+    export NUMBA_DPPY_DEBUGINFO=1
+    export NUMBA_OPT=1
```
akharche, line 18: Why we use NUMBA_OPT=1 every time, but recommend users to debug on no optimization level? May be it better to use "0" level?

docs/user_guides/debugging/backtrace.rst
```diff
@@ -0,0 +1,44 @@
...
+    gdb-oneapi -q --args python simple_dppy_func.py
+
+Next, we can see the call stack from the kernel and from the nested function:
```
akharche, line 21:
```suggestion
Next we see the call stack from the kernel and the nested function:
```

PokhodenkoSA (approved):

PR #466, opened 2021-06-18
Add assertion in @njit tests to ensure offloading

reazulhoque:
This PR makes sure we are asserting offload to SYCL device in the case of auto-offload.

During implementation it was discovered that `parfor` is not generated for `numoy.logical_not` and it was removed from testing as we can not offload regions that are not converted to `parfor` nodes.


diptorupd:
The changes here look good and needed. Thanks!

The question I had was why is `testing.py` under `numba_dppy`? Are those functions used any place outside of tests? If not so, the `testing.py` module should be moved into the tests directory as a helper module.

reazulhoque:
@diptorupd I have removed testing.py and moved it's content in numba_dppy/tests/_helper.py

numba_dppy/tests/__init__.py
```diff
@@ -12,11 +12,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""
```
diptorupd, line 15: 👍

numba_dppy/tests/__init__.py
```diff
@@ -12,11 +12,5 @@
...
-from numba.testing import SerialSuite
-from numba.testing import load_testsuite
-from os.path import dirname, join
-"""
-
 import numba_dppy
 from numba_dppy import config
```
diptorupd, line 16: Most probably these two imports are not needed. @reazulhoque can you try removing them and see if it works?

diptorupd (approved):
LGTM barring the small change in tests.__init__.py. Thanks!

PokhodenkoSA (approved):

PR #475, opened 2021-06-23
Update docs about breakpoints and backtrace

1e-to:
Update docs about breakpoints and backtrace

### DIFF
--- a/docs/user_guides/debugging/backtrace.rst
+++ b/docs/user_guides/debugging/backtrace.rst
@@ -12,37 +12,11 @@ Let's consider the work of the command ``backtrace`` in the following example ``
     Known issues:
         - The first line of the kernel and functions is hit twice. See the :ref:`single_stepping`.
 
-Run GDB debugger:
+Below are examples showing the backtrace for the kernel and for the nested function.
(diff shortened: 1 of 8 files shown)


docs/user_guides/debugging/backtrace.rst
```diff
@@ -12,37 +12,11 @@ Let's consider the work of the command ``backtrace`` in the following example ``
...
-    15          c_in_kernel[i] = func_sum(a_in_kernel[i], b_in_kernel[i])
-    (gdb) step
-    __main__::func_sum () at simple_dppy_func.py:8
-    8          result = a_in_func + b_in_func
-    (gdb) backtrace
-    #0  __main__::func_sum () at simple_dppy_func.py:8
-    #1  __main__::kernel_sum () at simple_dppy_func.py:15
-    (gdb) continue
-    Continuing.
+Below are examples showing the backtrace for the kernel and for the nested function.
```
diptorupd, line 15: How about:

> The section presents two examples of using GDB to generate backtrace from a ``numa_dppy.kernel`` function. The first example presents the case where the kernel function does not invoke any other function. The second example presents the case where the kernel function invokes a ``numba_dppy.func``.

After this small introduction you can then present the two cases one after the other.

> **Example 1** _add the code example followed by back trace._
> **Example 2** _add the code example followed by back trace._

PokhodenkoSA (approved):

PR #476, opened 2021-06-23
Raise error when users are trying to return from kernel

reazulhoque:
`@dppy.kernel` does not allow users to return anything. This PR adds checks to make sure this behavior is enforced and an useful error message is generated when users are trying to return from `dppy.kernel`

### DIFF
--- a/numba_dppy/compiler.py
+++ b/numba_dppy/compiler.py
@@ -91,6 +91,25 @@ class DPPYCompiler(CompilerBase):
 
 
 def compile_with_dppy(pyfunc, return_type, args, is_kernel, debug=None):
+    """
+    Compiles with Numba_dppy's pipeline and returns the compiled result.
+
+    Args:
+        pyfunc: The Python function to be compiled.
+        return_type: The Numba type of the return value.
+        args: The list of arguments sent to the Python function.
+        is_kernel (bool): Indicates whether the function is decorated
+            with @dppy.kernel or not.
+        debug (bool): Flag to turn debug mode ON/OFF.
+
+    Returns:
+        cres: Compiled result.
+
+    Raises:
+        TypeError: @dppy.kernel does not allow users to return any
+            value. TypeError is raised when users do.
+
+    """
     # First compilation will trigger the initialization of the OpenCL backend.
     from .descriptor import dppy_target
 
@@ -159,7 +178,9 @@ def compile_kernel(sycl_queue, pyfunc, args, access_types, debug=None):
         # We expect the sycl_queue to be provided when this function is called
         raise ValueError("SYCL queue is required for compiling a kernel")
 
-    cres = compile_with_dppy(pyfunc, None, args, True, debug=debug)
+    cres = compile_with_dppy(
+        pyfunc=pyfunc, return_type=None, args=args, is_kernel=True, debug=debug
+    )
     func = cres.library.get_function(cres.fndesc.llvm_func_name)
     kernel = cres.target_context.prepare_ocl_kernel(func, cres.signature.args)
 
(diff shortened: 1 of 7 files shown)


numba_dppy/compiler.py
```diff
@@ -154,7 +159,7 @@ def compile_kernel(sycl_queue, pyfunc, args, access_types, debug=None):
         # We expect the sycl_queue to be provided when this function is called
         raise ValueError("SYCL queue is required for compiling a kernel")
 
-    cres = compile_with_dppy(pyfunc, None, args, debug=debug)
+    cres = compile_with_dppy(pyfunc, None, args, True, debug=debug)
```
diptorupd, line 162: Use kwargs (return_type=None, is_kernel=True) as that makes reading the code much easier and understanding what each argument does.

numba_dppy/compiler.py
```diff
@@ -89,7 +90,7 @@ def define_pipelines(self):
         return pms
 
 
-def compile_with_dppy(pyfunc, return_type, args, debug=None):
+def compile_with_dppy(pyfunc, return_type, args, is_kernel, debug=None):
```
diptorupd, line 93: Please add a docstring and mention that the function can raise a `TypeError`.

numba_dppy/decorators.py
```diff
@@ -60,9 +60,7 @@ def _kernel_jit(signature, debug, access_types):
         ]
     )
 
-    if restype is not None and restype != types.void:
-        msg = "DPPY kernel must have void return type but got {restype}"
-        raise TypeError(msg.format(restype=restype))
+    assert_no_return(rettype)
```
diptorupd, line 63: A comment will be useful.

diptorupd (approved):
Please address the code style comments and merge.

numba_dppy/tests/kernel_tests/test_return_type.py
```diff
@@ -0,0 +1,47 @@
...
+    if skip_test(offload_device):
+        pytest.skip()
+
+    a = np.array(np.random.random(122), np.int32)
+
+    with pytest.raises(Exception):
+        kernel = dppy.kernel(sig)(f)
+
+        with dpctl.device_context(offload_device):
+            kernel[a.size, dppy.DEFAULT_LOCAL_SIZE](a)
```
PokhodenkoSA, line 47: 1. Exception should be concrete. Just catching any exception could hide a problem.
2. I think `dppy.kernel(sig)(f)` does not throw exception. So `with pytest.raises(Exception):` should wrap only function call.

numba_dppy/decorators.py
PokhodenkoSA: If comment needed then function should have better name.

numba_dppy/compiler.py
reazulhoque: I do not want to use kwargs here for the following reasons:
1. Keyword arguments are optional unless we make all arguments keyword arguments and use techniques described here (https://www.vegardstikbakke.com/python-keyword-only/). Please let me know if this is a wrong assumption.
2. I do not want the function to assume a default value as I want future developers to explicitly tell this function whether this is a kernel or not. In case keyword arguments are optional, future developers might not provide it and the default value will be assumed.

Making the code readable is the only positive aspect I see if my understanding is correct. Please correct me if I am under wrong assumptions.

numba_dppy/tests/kernel_tests/test_return_type.py
reazulhoque: 1. I agree but this function raises `TypeError` which is a generic error and is suseptible to your original comment. In this case should we make our own error type to make it actually concrete? Is this scalable?
2. It does and hence the wrap. If signature is provided the `exception` would be raised at the time this function is going through the decorator.

PR #498, opened 2021-07-14
Remove `llvmdev` from runtime dependecies

reazulhoque:
This PR updates the runtime dependencies specified in the conda recipe.

1. Remove `llvmdev` as we do not use `opt` to generate LLVM bitcode.
2. Introduce `packaging` as it is being used in  `numba_dppy/config.py`.

@PokhodenkoSA Do we need to use packaging [here](https://github.com/IntelPython/numba-dppy/blob/main/numba_dppy/config.py#L17) or can it be accomplished somehow else?

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -28,7 +28,7 @@ requirements:
         - dpctl 0.8.*
         - spirv-tools
         - llvm-spirv
-        - llvmdev
+        - packaging
         - dpnp >=0.6*,<0.7*  # [linux]
 
 test:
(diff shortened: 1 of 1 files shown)


PokhodenkoSA:
@reazulhoque please provide PR where it started:
> 1. ... we do not use `opt` to generate LLVM bitcode.

From this:
> 2. Introduce `packaging` as it is being used in  `numba_dppy/config.py`.

As I understand `llvmdev` depends on `packaging`. So now `llvmdev` is not used but `packaging` is needed. Correct?

If yes then I think the changes are good.

PokhodenkoSA (approved):

reazulhoque:
@PokhodenkoSA I do not see `packaging` as a dependency even when we have `llvmdev` as a runtime dependency. Both the points are applicable to this PR.

PR #515, opened 2021-08-17
Update go version

reazulhoque:
The latest update in go version is causing license step in CI to fail. This is an attempt to setup a version of go that will not fail to install license  using github action.

### DIFF
--- a/.github/workflows/license.yml
+++ b/.github/workflows/license.yml
@@ -12,6 +12,8 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-go@v2
+      with:
+          go-version: '1.16'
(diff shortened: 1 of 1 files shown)


oleksandr-pavlyk:
License workflow is green, so all is good.

oleksandr-pavlyk (approved):
Just out of the curiosity, is there a reason we do not use the latest one?

reazulhoque:
I did not originally set this up. Without the version specified I am not sure which version of GO runs. I got concerned about this as our license workflow is failing for the latest PRs (https://github.com/IntelPython/numba-dppy/pull/493). While looking for a solution someone suggested upgrading to the version proposed solves this issue. Hence I am using that version.

PR #529, opened 2021-09-01
Update/dpctl memcpy async

reazulhoque:
This updates API changes in dpctl for memcpy. https://github.com/IntelPython/dpctl/pull/557


conda-recipe/meta.yaml
```diff
@@ -19,13 +19,13 @@ requirements:
         - setuptools
         - cython
         - numba 0.54*
-        - dpctl >=0.9*
+        - dpctl >=0.9.1.dev*
```
oleksandr-pavlyk, line 22: This will presently make it break, since no such package is yet available on the channels.

PokhodenkoSA: @oleksandr-pavlyk 
If API changed then dpctl version should be 0.10.0dev0.
0.9.1 is a patch version.
Right?

oleksandr-pavlyk: To play nice with multi-versioning on Linux - yes.

numba_dppy/dpnp_glue/dpnp_array_creations_impl.py
```diff
@@ -31,15 +31,21 @@ def common_impl(a, b, out, dpnp_func, PRINT_DEBUG):
     sycl_queue = dpctl_functions.get_current_queue()
 
     b_usm = dpctl_functions.malloc_shared(b.size * b.itemsize, sycl_queue)
-    dpctl_functions.queue_memcpy(sycl_queue, b_usm, b.ctypes, b.size * b.itemsize)
+    event = dpctl_functions.queue_memcpy(
```
oleksandr-pavlyk, line 34: Wouldn't it be better to create a function:

```python
def synchronous_memcpy(q, dst, src, size):
     event = dpctl_functions.queue_memcpy(q, dst, src, size)
     dpctl_functions.event_wait(event)
     dpctl_functions.event_delete(event)
```

and use it throughout?

conda-recipe/meta.yaml
```diff
@@ -19,13 +19,13 @@ requirements:
...
         - dpnp >=0.7*  # [linux]
         - wheel
     run:
         - python
         - numba 0.54*
-        - dpctl >=0.9*
+        - dpctl >=0.9.1.dev*
```
PokhodenkoSA, line 28: @reazulhoque It is not necessary to use `dev` here. Just `>=0.9.1*` 
```suggestion
        - dpctl >=0.9.1*
```

PokhodenkoSA:
```suggestion
        - dpctl >=0.10.0*
```

PokhodenkoSA:
Waiting for dpctl 0.10.0dev0 uploaded to channels.

numba_dppy/dpnp_glue/dpnp_array_creations_impl.py
oleksandr-pavlyk: Opinions?

PokhodenkoSA: About dpctl API I would like to propose to try to avoid braking changes:
```
def queue_memcpy(..., wait=True): 
    ...
    return event

def queue_memcpy_async(...):
    return queue_memcpy(..., wait=False)
```

PokhodenkoSA (approved):
I think we can merge it.

PR #546, opened 2021-09-13
Add pylint

PokhodenkoSA:


diptorupd:
Adding `pylint` support is a good goal, but not a priority at this point. I am opening an issue ticket to evaluate pylint support in dpex as a future task (#799).

As the PR is a draft, closing it

PR #566, opened 2021-09-22
Test numba and numba-dppy API with GDB

### LINKED ISSUE #528: Test debugging for numba-dppy and numba side-by-side
- [ ] Check that numba has the same behavior with local variables as Numba-DPPy.
  - [ ] We need to reproduce the same issues with Vanilla Numba.

In tests we should have scripts, which compiles with Vanilla Numba and with Numba-DPPy.
E.g. an @njit kernel offloaded to CPU.
Same issues e.g. with variable values shows with both Vanilla Numba and Numba-DPPy.
It would be a good starting point, a script to follow when and how the issue will get fixed.

PokhodenkoSA:
Fixes #528.
Closes #564.
Closes #548.


numba_dppy/tests/test_debug_dppy_numba.py
```diff
@@ -0,0 +1,89 @@
...
+    @staticmethod
+    def script_path(script):
+        import os
+
+        return os.path.join(
+            os.path.dirname(os.path.abspath(numba_dppy.__file__)),
+            "examples",
+            "debug",
+            script,
+        )
```
akharche, line 79:
```suggestion
        import pathlib

        return pathlib.Path(numba_dppy.__file__).parents[0] / 'examples/debug/'
```

1e-to (approved):

PR #587, opened 2021-10-15
Run GDB tests [internal CI]

vlad-perevezentsev:
This PPR includes PR #579  with the addition and modification of tests for GDB for new `Numba` debugging features [SAT - 4286].


diptorupd:
I am closing the PR since it is by now two years old and hopelessly bitrotten. If in future, the effort is to be revived we can refer to how the GDB setup is done in our public github CI.

PR #622, opened 2021-11-09
Update clang to icx

reazulhoque:
As title.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -102,9 +102,9 @@ def _get_cmdclass():
 
 def spirv_compile():
     if IS_LIN:
-        compiler = "clang"
+        compiler = "icx"
     if IS_WIN:
-        compiler = "clang.exe"
+        compiler = "icx.exe"
 
     clang_args = [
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -102,9 +102,9 @@ def _get_cmdclass():
 
 def spirv_compile():
     if IS_LIN:
-        compiler = "clang"
+        compiler = "icx"
     if IS_WIN:
-        compiler = "clang.exe"
+        compiler = "icx.exe"
```
diptorupd, line 107: @reazulhoque is the Windows executable called `icx.exe`? CI seems to be dead, so is there any other way to verify?

diptorupd (approved):

setup.py
reazulhoque: Verified, it is there.

PokhodenkoSA:
@reazulhoque please, check how this PR affected internal CI. @1e-to said CI fails because of it.

PR #627, opened 2021-11-15
Use llvm-spirv from bin-llvm during build on Windows

PokhodenkoSA:


setup.py
```diff
@@ -104,7 +104,7 @@ def spirv_compile():
     if IS_LIN:
         compiler = "icx"
     if IS_WIN:
-        compiler = "icx.exe"
+        compiler = "clang.exe"
```
PokhodenkoSA, line 107: With `icx` it fallback to `clang-cl` which does not understand all our input arguments.

conda-recipe/bld.bat
```diff
@@ -1,3 +1,7 @@
+@REM new llvm-spirv location
+@REM starting from dpcpp_impl_win-64=2022.0.0=intel_3638 location is env\Library\bin-llvm
+set PATH=%LIBRARY_PREFIX%\bin-llvm;%PATH%
```
oleksandr-pavlyk, line 3: `%PATH%` is only modified during building. Does `numba_dppy` also need to know where to look for llvm-spirv at runtime?

PokhodenkoSA: At runtime it uses package `llvm-spirv`.  We decided to not install compiler in runtime. But this package is public. Compiler team does not provide separate package for `llvm-spirv` yet.
Possibly there is inconsistency between llvm-spirv received from `dpcpp` and `llvm-spirv`. @reazulhoque mention some bugs related to this. The second solution could be to install `dpcpp` package in run time.

PR #631, opened 2021-11-19
Change black to only allow 80 char lines. Reformat sources.

diptorupd:
Changes the black configuration to use 80 char wide lines to be consistent with `dpctl` and PEP-8. All sources are reformatted. I will open a different PR to ignore this commit from git-blame.


diptorupd:
@PokhodenkoSA @reazulhoque I am merging it as CI has passed and it is a formatting only change.

PR #637, opened 2021-11-22
Move coverage.py configs into pyproject.toml

diptorupd:
We do not need a separate config file for `coverage.py` and the config options can be included inside `pyproject.toml`.


diptorupd:
Merging as coverage generation worked as expected and no other code was touched.

PR #645, opened 2021-11-25
prep for dpctl/include reorg (IntelPython/dpctl#685)

oleksandr-pavlyk:
Use `__has_include` to work with both layouts.


PokhodenkoSA:
Used in #669.

PR #646, opened 2021-11-26
Fix skipping for Level Zero device for "with context" test

1e-to:
Missed check on existing device
Without level-zero:
```python
FAILED numba_dppy/tests/test_with_context.py::test_dpctl_device_context_affects_numba_pipeline[device_context0-level_zero:gpu:0] - V...
FAILED numba_dppy/tests/test_with_context.py::test_dpctl_device_context_affects_numba_pipeline[device_context1-level_zero:gpu:0] - V...
```

### DIFF
--- a/numba_dppy/tests/test_with_context.py
+++ b/numba_dppy/tests/test_with_context.py
@@ -30,10 +30,13 @@ skip_no_gpu = pytest.mark.skipif(
 skip_no_cpu = pytest.mark.skipif(
     not _helper.has_cpu_queues(), reason="No CPU platforms available"
 )
-
+skip_no_level_zero = pytest.mark.skipif(
+    not _helper.has_gpu_queues("level_zero"),
+    reason="No level-zero GPU platforms available",
+)
 
 filter_strings = [
-    pytest.param("level_zero:gpu:0", marks=skip_no_gpu),
+    pytest.param("level_zero:gpu:0", marks=skip_no_level_zero),
     pytest.param("opencl:gpu:0", marks=skip_no_gpu),
     pytest.param("opencl:cpu:0", marks=skip_no_cpu),
 ]
--- a/numba_dppy/tests/test_with_context.py
+++ b/numba_dppy/tests/test_with_context.py
@@ -25,10 +25,12 @@ from . import _helper
 from ._helper import assert_auto_offloading
 
 skip_no_gpu = pytest.mark.skipif(
-    not _helper.has_gpu_queues(), reason="No GPU platforms available"
+    not _helper.has_gpu_queues("opencl"),
+    reason="No opencl GPU platforms available",
 )
 skip_no_cpu = pytest.mark.skipif(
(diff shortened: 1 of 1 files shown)


numba_dppy/tests/test_with_context.py
```diff
@@ -49,6 +49,8 @@ def func(a, b):
     a = np.ones((64), dtype=np.int32)
     b = np.ones((64), dtype=np.int32)
 
+    if skip_test(filter_str):
+        pytest.skip()
```
PokhodenkoSA, line 53: The problem is not here. In line 38 `pytest.param("level_zero:gpu:0", marks=skip_no_gpu)` should be skipped if GPU is not available. So it should be skipped if L0 is not available.

PokhodenkoSA (changes requested):

numba_dppy/tests/test_with_context.py
```diff
@@ -24,18 +24,28 @@
...
-    pytest.param("level_zero:gpu:0", marks=skip_no_gpu),
-    pytest.param("opencl:gpu:0", marks=skip_no_gpu),
-    pytest.param("opencl:cpu:0", marks=skip_no_cpu),
+    pytest.param(
+        "level_zero:gpu:0",
+        marks=pytest.mark.skipif(
+            _helper.skip_test("level_zero:gpu:0"),
+            reason="No GPU platforms available",
+        ),
+    ),
```
PokhodenkoSA, line 34: 3 indentations. I think it is too much.

numba_dppy/tests/test_with_context.py
```diff
@@ -24,18 +24,28 @@
...
+    pytest.param(
+        "opencl:gpu:0",
+        marks=pytest.mark.skipif(
+            _helper.skip_test("opencl:gpu:0"),
```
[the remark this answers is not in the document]
1e-to: What about rewriting the skip_test function in such a way that all device checks are included in it (is there a gpu, cpu, opencl, level-zero) and return the message why test fail. And accordingly change the skip to xfail?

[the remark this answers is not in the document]
PokhodenkoSA: > all device checks are included in it

This functions should be named - `check_device`, not `skip_test`.

> and return the message why test fail.

It seems that function will have second responsibility.

I would prefer one function with simple responsibility - return pytest mark for the device.

numba_dppy/tests/test_with_context.py
```diff
@@ -30,10 +30,13 @@
 skip_no_cpu = pytest.mark.skipif(
     not _helper.has_cpu_queues(), reason="No CPU platforms available"
 )
-
+skip_no_level_zero = pytest.mark.skipif(
+    not _helper.has_gpu_queues("level_zero"),
```
PokhodenkoSA, line 34: Please, pass also "opencl" for `skip_no_gpu` to make it obvious.

PokhodenkoSA (approved):

PR #649, opened 2021-11-29
Use `llvm-spirv` from `dpcpp` compiler package by default

PokhodenkoSA:
If `dpcpp` compiler package installed then `numba-dppy` uses `llvm-spirv` from it by default.
User could provide `LLVM_SPIRV_ROOT` to change it.

No need for defining 2 variables like here:
https://github.com/IntelPython/numba-dppy/blob/aa67724fc9a3175138264f36b034b9f8dc4a2efd/numba_dppy/examples/atomic_op.py#L33

- [x] No `llvm-spirv` package dependency
- [x] `dpcpp` package run-time dependency
- [x] Search `llvm-spirv` as `bin/dpcpp/../../bin-llvm/llvm-spirv`
- [x] Fallback to `llvm-spirv` if `dpcpp` and `LLVM_SPIRV_ROOT` don't exist
- [x] Print path to llvm-spirv for debugging

### DIFF
--- a/numba_dppy/spirv_generator.py
+++ b/numba_dppy/spirv_generator.py
@@ -85,27 +85,34 @@ class CmdLine:
         if config.DEBUG:
             llvm_spirv_flags.append("--spirv-debug-info-version=ocl-100")
 
-        if config.LLVM_SPIRV_ROOT:
-            llvm_spirv_tool = shutil.which(
-                "llvm-spirv", path=config.LLVM_SPIRV_ROOT
-            )
+        llvm_spirv_tool = self._llvm_spirv()
 
-        if not llvm_spirv_tool:
+        if config.DEBUG:
+            print(f"Use llvm-spirv: {llvm_spirv_tool}")
+
+        check_call([llvm_spirv_tool, *llvm_spirv_args, "-o", opath, ipath])
+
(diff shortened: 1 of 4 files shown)


numba_dppy/spirv_generator.py
```diff
@@ -84,18 +85,21 @@ def generate(self, llvm_spirv_args, ipath, opath):
         if config.DEBUG:
             llvm_spirv_flags.append("--spirv-debug-info-version=ocl-100")
 
-        llvm_spirv_tool = "llvm-spirv"
-        if config.NATIVE_FP_ATOMICS == 1:
-            llvm_spirv_root = config.LLVM_SPIRV_ROOT
+        # use llvm-spirv from dpcpp package.
+        # assume dpcpp from .../bin folder.
+        # assume llvm-spirf from .../bin-llvm folder.
```
reazulhoque, line 90: Typo

reazulhoque (approved):

numba_dppy/spirv_generator.py
PokhodenkoSA:
```suggestion
        # assume llvm-spirv from .../bin-llvm folder.
```

PR #652, opened 2021-11-30
Redesign helper functions

1e-to:
Proposal to change functions 'has_gpu_queues' and 'has_cpu_queues' to three new: has_opencl_gpu, has_opencl_cpu, has_level_zero

### DIFF
--- a/numba_dppy/__init__.py
+++ b/numba_dppy/__init__.py
@@ -208,10 +208,10 @@ Complete Example using @dppy.kernel:
         b = np.array(np.random.random(N), dtype=np.float32)
         c = np.ones_like(a)
 
-        if dpctl.has_opencl_gpu():
+        if dpctl.has_gpu_queues():
             with dpctl.device_context("opencl:gpu") as gpu_queue:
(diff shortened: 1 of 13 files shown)


numba_dppy/__init__.py
```diff
@@ -208,10 +208,10 @@ def main():
         b = np.array(np.random.random(N), dtype=np.float32)
         c = np.ones_like(a)
 
-        if dpctl.has_gpu_queues():
+        if dpctl.has_opencl_gpu():
```
PokhodenkoSA, line 211: Dose `dpctl` have `has_opencl_gpu` ?

diptorupd: No. What is the intent behind these changes? The whole docstring needs to be updated. It is on my long list of todo

1e-to: Yes, you're right, there is no such method. In general, this example itself is outdated, because the method `has_gpu_queues` no longer exists.

PokhodenkoSA (approved):

PR #659, opened 2021-12-02
Move `dpcpp/llvm-spirv` from runtime to testing dependency

PokhodenkoSA:


reazulhoque (approved):
Pending the split test when users do not provide flag to generate the native atomic operations.

PR #669, opened 2021-12-08
Support dpctl 0.12

PokhodenkoSA:
- [x] Compatible with dpctl 0.11
- [x] Build with dpctl 0.11.0
- [x] Test with dpctl 0.11 and 0.12

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -19,13 +19,13 @@ requirements:
         - setuptools
         - cython
         - numba 0.54*|0.55*
-        - dpctl 0.12*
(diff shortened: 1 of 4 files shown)


oleksandr-pavlyk:
Please see https://github.com/IntelPython/dpctl/pull/685#issuecomment-977897586 for another way of allowing compatibility with both 0.11 and 0.12 until 0.12 becomes commonplace and the preprocessor macros can be removed.

I even opened a PR #645 in numba-dppy to add that.  @diptorupd was not in favor.

PokhodenkoSA:
@oleksandr-pavlyk 
Thanks for solution. Applied solution from #645.

PR #672, opened 2021-12-09
Use filter_str to skip tests on missing devices

1e-to:
Use **filter_str** to skip tests on missing devices


PokhodenkoSA:
@1e-to please, provide description. See https://github.com/IntelPython/numba-dppy/wiki/Definition-of-DONE.

PokhodenkoSA (approved):
I would prefer 2 steps - 1. use new decorator, 2. replace `offload_device` variable with `filter_str`

PR #685, opened 2022-01-11
Add mechanics to skip tests on missing devices by env var

1e-to:
Add mechanics to skip tests on missing devices by env var


PokhodenkoSA (approved):
LGTM

numba_dppy/tests/_helper.py
```diff
@@ -88,15 +88,15 @@ def is_windows():
 skip_windows = pytest.mark.skipif(is_windows(), reason="Skip on Windows")
 
 skip_no_opencl_gpu = pytest.mark.skipif(
-    not has_opencl_gpu(),
+    not has_opencl_gpu() and config.TESTING_SKIP_NO_OPENCL_GPU,
```
chudur-budur, line 91: `has_opencl_gpu()` is supposed to skip a test when there is no opencl GPU already, so why do we need this? Not clear from the PR description.

numba_dppy/tests/_helper.py
```diff
@@ -88,15 +88,15 @@ def is_windows():
...
     reason="No opencl GPU platforms available",
 )
 skip_no_opencl_cpu = pytest.mark.skipif(
-    not has_opencl_cpu(),
+    not has_opencl_cpu() and config.TESTING_SKIP_NO_OPENCL_CPU,
```
chudur-budur, line 95: `has_opencl_cpu()` is supposed to skip a test when there is no opencl CPU already, why do we need this? Not clear from the PR description.

numba_dppy/tests/_helper.py
```diff
@@ -88,15 +88,15 @@ def is_windows():
...
     reason="No opencl CPU platforms available",
 )
 skip_no_level_zero_gpu = pytest.mark.skipif(
-    not has_level_zero(),
+    not has_level_zero() and config.TESTING_SKIP_NO_LEVEL_ZERO_GPU,
```
chudur-budur, line 99: `has_level_zero()` is supposed to skip a test when there is no level zero device already, why do we need this? Not clear from the PR description.

numba_dppy/tests/_helper.py
```diff
@@ -123,6 +123,34 @@ def is_windows():
     pytest.param("level_zero:gpu:0", marks=skip_no_level_zero_gpu),
 ]
 
+skip_opencl_gpu_not_default_device = pytest.mark.skipif(
+    "opencl:gpu" not in dpctl.get_current_queue().sycl_device.filter_string,
```
chudur-budur, line 127: If `has_opencl_gpu()` is already returning `False`, then performing this check has no point, it's going to be `False` anyway. Not clear what is going on here.

numba_dppy/tests/_helper.py
```diff
@@ -123,6 +123,34 @@ def is_windows():
...
+    reason="Works only with default device",
+)
+skip_opencl_cpu_not_default_device = pytest.mark.skipif(
+    "opencl:cpu" not in dpctl.get_current_queue().sycl_device.filter_string,
```
chudur-budur, line 131: If `has_opencl_cpu()` is already returning `False`, then performing this check has no point, it's going to be `False` anyway. Not clear what is going on here.

numba_dppy/tests/_helper.py
```diff
@@ -123,6 +123,34 @@ def is_windows():
...
+    reason="Works only with default device",
+)
+skip_level_zero_gpu_not_default_device = pytest.mark.skipif(
+    "level_zero:gpu" not in dpctl.get_current_queue().sycl_device.filter_string,
```
chudur-budur, line 135: If `has_level_zero()` is already returning `False`, then performing this check has no point, it's going to be `False` anyway. Not clear what is going on here.

numba_dppy/tests/_helper.py
```diff
@@ -123,6 +123,34 @@ def is_windows():
...
+    reason="Works only with default device",
+)
+
+filter_strings_not_default_device = [
```
chudur-budur, line 139: If all the previous modifications do not hold, this list becomes redundant.

numba_dppy/tests/integration/test_dpnp_interop.py
```diff
@@ -17,7 +17,7 @@
 import pytest
 
 import numba_dppy as dpex
-from numba_dppy.tests._helper import filter_strings
+from numba_dppy.tests._helper import filter_strings_not_default_device
```
chudur-budur, line 20: Therefore, we don't need this anymore.

chudur-budur (commented):
This PR is not very clear about why we need this. It only makes sense if `has_opencl_gpu()`, `has_opencl_gpu()` or `has_level_zero()` functions are buggy (i.e. they are not returning correct values at correct situation). If that is the case, I think we should open an issue explaining what is going on with those functions.

Also the changes need to be done on the latest main branch, for example we don't have `numba_dppy` anymore. Could you also check "Allow edits by maintainers"? 

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

diptorupd:
I am closing this PR. We will be redesigning the test suite completely post 0.20 and some these concerns will be addressed in a different way.

PR #694, opened 2022-02-02
Register global functions

### LINKED ISSUE #693: `get_global_id` does not work if function is directly exported
If I try the following code

```
import dpctl
import base_l2_distance
import math

from numba_dppy import kernel, get_global_id, atomic, DEFAULT_LOCAL_SIZE
atomic_add = atomic.add

@kernel(access_types={"read_only": ["a", "b"], "write_only": ["c"]})
def l2_distance_kernel(a, b, c):
    i = get_global_id(0)
    j = get_global_id(1)
    sub = a[i, j] - b[i, j]
    sq = sub**2
    atomic_add(c, 0, sq)

def l2_distance(a, b, distance):
    with dpctl.device_context(base_l2_distance.get_device_selector()):
        l2_distance_kernel[(a.shape[0], a.shape[1]), DEFAULT_LOCAL_SIZE](a, b, distance)

    return math.sqrt(distance)

```
the following error is risen during typeinfer

```
  File "/nfs/site/home/mkolpako/code/miniconda3/envs/gpu_dev/lib/python3.7/site-packages/numba/core/typeinfer.py", line 1466, in resolve_value_type
    raise TypingError(msg, loc=inst.loc)
numba.core.errors.TypingError: Failed in dppy_nopython mode pipeline (step: nopython frontend)
Untyped global name 'get_global_id': Cannot determine Numba type of <class 'function'>

File "l2_distance_kernel.py", line 59:
def l2_distance_kernel(a, b, c):
    i = get_global_id(0)
```

Everything works if I use `numba_dppy.get_global_id` notation

PokhodenkoSA:
Closes #693.

The dpex kernel API intrinsic functions such as `dpex.get_global_id` always need to be called with the fully qualified name including the module. As described in issue #693, these functions cannot be directly called inside a kernel function.

The PR adds the intrinsic functions in dpex to the `globals` dictionary to allow the usage without a module name specifier.

### DIFF
--- a/numba_dpex/tests/ocl/__init__.py
+++ b/numba_dpex/tests/ocl/__init__.py
@@ -1,0 +1,13 @@
+# Copyright 2021 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
--- a/numba_dpex/tests/ocl/__init__.py
+++ b/numba_dpex/tests/ocl/__init__.py
@@ -1,13 +1,3 @@
-# Copyright 2021 Intel Corporation
+# Copyright 2021 - 2022 Intel Corporation
 #
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#      http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
+# SPDX-License-Identifier: Apache-2.0
--- a/numba_dpex/tests/ocl/test_ocldecl.py
+++ b/numba_dpex/tests/ocl/test_ocldecl.py
@@ -1,0 +1,43 @@
+# Copyright 2021 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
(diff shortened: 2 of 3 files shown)


numba_dppy/ocl/ocldecl.py
```diff
@@ -35,66 +35,32 @@
...
-class Ocl_mem_fence(ConcreteTemplate):
-    key = dppy.mem_fence
-    cases = [signature(types.void, types.uint32)]
-
-
-@intrinsic
-class Ocl_sub_group_barrier(ConcreteTemplate):
-    key = dppy.sub_group_barrier
+    globals()[name] = intrinsic(cls)
+    intrinsic_global(function, types.Function(cls))
```
PokhodenkoSA, line 51: The only new thing is here. 
Other changes are only for refactoring and testing.

PokhodenkoSA:
It breaks a lot of tests in internal CI. Need more investigations.

diptorupd:
@mingjie-intel I made some fixes to Sergey's patch and the tests should work now. I have updated the PR description to describe what the change does.

There are two issues I have with the PR and the issue #693:

1. The PR does not address all `dpex` intrinsics (e.g. `dpex.atomic.add`) will not work yet. We can fix it by doing the same trick of adding the function to `globals`.
2. A more philosophical issue is should we even allow intrinsics to be called without the module name specifier. I looked at what `numba.cuda` does and the too only allow calling these functions via the module name specifier. Moreover, adding these functions to globals does not seem right to be.

What are your thoughts? 

+ @oleksandr-pavlyk + @chudur-budur

numba_dpex/tests/ocl/test_ocldecl.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2021 Intel Corporation
```
oleksandr-pavlyk, line 1:
```suggestion
# Copyright 2022 Intel Corporation
```

numba_dpex/tests/ocl/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2021 Intel Corporation
```
oleksandr-pavlyk, line 1:
```suggestion
# Copyright 2022 Intel Corporation
```

mingjie-intel:
In general, I feel making global is not good code practices. Unless there is strong push from user side, I would recommend that we close this PR. @oleksandr-pavlyk   @chudur-budur What do you think?

numba_dpex/tests/ocl/test_ocldecl.py
diptorupd: done

numba_dpex/tests/ocl/__init__.py
diptorupd: done

diptorupd:
> In general, I feel making global is not good code practices. Unless there is strong push from user side, I would recommend that we close this PR. @oleksandr-pavlyk @chudur-budur What do you think?

I tend to agree. Letting users write `get_global_id` instead of `numba_dpex.get_global_id` is a minor syntactic preference IMO. If we want to be pedantic perhaps we can go the route of `numba.cuda` and introduce a `sycl` module and change all of these functions to `sycl` module.

```
from numba_dpex import sycl

@sycl.kernel
def vecadd(a,b,c):
   i = sycl.get_global_id(0)
   a[i] = b[i] + c[i]
```

Having the module qualifier is much better in terms of overall readability and letting users know that the code is specific to the numba_dpex kernel API and will not work outside of the kernel function or without the kernel decorator.

diptorupd:
Based on feedback, I am closing the PR and the related issue as WONTFIX.

PR #698, opened 2022-02-03
Fix dpctl version in recipe 

1e-to:
Fix dpctl version in recipe

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -26,7 +26,7 @@ requirements:
     run:
         - python
         - numba 0.54*|0.55*
-        - dpctl 0.11*|0.12*
+        - dpctl > 0.11.1
         - spirv-tools
         - llvm-spirv 11.*
         - dpnp                     # [linux]


conda-recipe/meta.yaml
```diff
@@ -26,7 +26,7 @@ requirements:
     run:
         - python
         - numba 0.54*|0.55*
-        - dpctl 0.11*|0.12*
+        - dpctl >0.11.1
```
PokhodenkoSA, line 29: Why remove `0.12*`?

conda-recipe/conda_build_config.yaml
```diff
@@ -6,7 +6,7 @@ cxx_compiler_version:       # [linux]
   - 7                       # [linux and x86_64]
 
 dpctl:
-  - 0.11.0
+  - 0.11.2
```
PokhodenkoSA, line 9: Is `0.11.2` available?

PokhodenkoSA: The policy is to build with first minor version (0.11.0) and test with any patch version (0.11.2). That is why 0.11.0.

PR #701, opened 2022-02-08
Fix public CI: Update package versions (mkl, dpctl, dpnp)

1e-to:
Update package versions (mkl, dpctl, dpnp) and versions in public CI

### DIFF
--- a/conda-recipe/conda_build_config.yaml
+++ b/conda-recipe/conda_build_config.yaml
@@ -6,13 +6,13 @@ cxx_compiler_version:       # [linux]
   - 7                       # [linux and x86_64]
 
 dpctl:
-  - 0.11.0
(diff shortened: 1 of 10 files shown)


conda-recipe/conda_build_config.yaml
```diff
@@ -6,13 +6,13 @@ cxx_compiler_version:       # [linux]
...
-  - 0.11.0
+  - 0.12.0dev3
 
 dpnp:
-  - 0.9.0dev0
+  - 0.9.0=*_0
 
 mkl:
-  - 2021.3.0
+  - 2022.0.1
```
PokhodenkoSA, line 15: If dpnp pins to mkl version then it is not necessary.

PR #705, opened 2022-02-10
Move `conda_build_config.yaml` to `scripts`

PokhodenkoSA:
Original contributor did not care to provide a description outlining the motivation for this change. 

I suspect this change of file location is to cooperate with internal CI for which this variant was not useful. Of course internal CI can override the `conda_build_config` by using `-m` option, but that's a lot of manual changes.


diptorupd:
too bitrotten at this point.

PR #727, opened 2022-03-24
Search `llvm-spirv` for `setup.py` in folder `bin-llvm`

PokhodenkoSA:
Previously you had to add `bin-llvm` to `PATH`: `export PATH=$CONDA_PREFIX/bin-llvm:$PATH`. See `conda-recipe`.
Now `setup.py` can find `llvm-spirv` using path to `dpcpp`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -125,13 +125,35 @@ def spirv_compile():
         "numba_dppy/ocl/atomics/atomic_ops.bc",
     ]
     spirv_args = [
-        "llvm-spirv",
+        _llvm_spirv(),
         "-o",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -134,6 +134,28 @@ def spirv_compile():
     subprocess.check_call(spirv_args, stderr=subprocess.STDOUT, shell=False)
 
 
+def _llvm_spirv():
+    """Return path to llvm-spirv executable."""
+    import shutil
+
+    result = None
+
+    if result is None:
```
akharche, line 143: Is this condition that is always met?
``result = None``

PR #729, opened 2022-03-28
Remove DPPY prefix for modules and functions in numba_dppy.core.passes.

diptorupd:
In preparation for the renaming from numba_dppy to numba_dpex, the PR renames the modules `dppy_lowerer` and `dppy_passes` to just `lowerer` and `passes`. The goal is to reduce the usage of `DPPY` or `numba_dppy` prefixes as much as possible to easy the name change.

### DIFF
--- a/numba_dppy/compiler.py
+++ b/numba_dppy/compiler.py
@@ -45,7 +45,7 @@ from numba_dppy.utils import (
 )
 
 from . import spirv_generator
-from .dppy_passbuilder import DPPYPassBuilder
+from .dppy_passbuilder import PassBuilder
 
 _NUMBA_DPPY_READ_ONLY = "read_only"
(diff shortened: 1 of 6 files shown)


PokhodenkoSA (approved):

numba_dppy/compiler.py
```diff
@@ -45,7 +45,7 @@
 )
 
 from . import spirv_generator
-from .dppy_passbuilder import DPPYPassBuilder
+from .dppy_passbuilder import PassBuilder
```
PokhodenkoSA, line 48: `dppy_passbuilder` also contains `dppy`.

diptorupd: Yes, I am creating a separate commit to rename everything outside of `numba_dppy.core`.

PR #772, opened 2022-09-08
Fixed subprocess in setup.py. It was causing setup to fail.

mingjie-intel:
To prevent timeouts of hefty subprocess calls, redirect subprocess's output to DEVNULL.

N/A


setup.py
```diff
@@ -130,9 +130,10 @@ def spirv_compile():
         "numba_dpex/ocl/atomics/atomic_ops.spir",
         "numba_dpex/ocl/atomics/atomic_ops.bc",
     ]
-    subprocess.check_call(clang_args, stderr=subprocess.STDOUT, shell=False)
-    subprocess.check_call(spirv_args, stderr=subprocess.STDOUT, shell=False)
-
+    print(clang_args)
```
oleksandr-pavlyk, line 133:
```suggestion
```

setup.py
```diff
@@ -130,9 +130,10 @@ def spirv_compile():
...
+    print(spirv_args)
```
oleksandr-pavlyk, line 134:
```suggestion
```

oleksandr-pavlyk (approved):
If we are not using neither stdout nor stderr, it is the right thing to do.

diptorupd:
I have not been able to replicate the issue for a while. I am closing the PR for now. We can revisit if the issue crops back.

PR #797, opened 2022-10-14
 Vendored llvm-spirv from dpcpp package.

mingjie-intel:
1. Remove dependency on llvm-spirv package.
    2. Use llvm-spirv from dpcpp package.


setup.py
```diff
@@ -130,8 +130,8 @@ def spirv_compile():
         "numba_dpex/ocl/atomics/atomic_ops.spir",
         "numba_dpex/ocl/atomics/atomic_ops.bc",
     ]
-    subprocess.check_call(clang_args, stderr=subprocess.STDOUT, shell=False)
```
diptorupd, line 123: Is this change still needed?

conda-recipe/meta.yaml
```diff
@@ -1,9 +1,20 @@
+{% set version = "2022.2.0" %}          # [linux]
```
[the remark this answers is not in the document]
diptorupd: @oleksandr-pavlyk explained the need, seems like cannot be avoided.

numba_dpex/spirv_generator.py
```diff
@@ -86,22 +86,12 @@ def generate(self, llvm_spirv_args, ipath, opath):
     def _llvm_spirv():
         """Return path to llvm-spirv executable."""
         result = None
+        try:
+            import llvm_spirv
+        except ImportError:
+            raise ImportError("Cannot import llvm-spirv")
```
diptorupd, line 92: A more descriptive error message is needed here. The message implies that it is something wrong with user env, may be they need to install llvm-spirv. 

We should be clear that it is an internal numba-dpex error.

diptorupd: I am adding a new `Exception` to dpex_exceptions called `UnreachableError`. The idea is borrowed from LLVM, it indicates that a supposedly impossible branch inside the compiler got executed. 

We can leave the code here as is and change it to an `UnreachableError` later. The idea being that this is an exception that should never be thrown.

setup.py
diptorupd: Based on discussions, I have reopened #772. Let us merge that separate PR rather then include the unrelated change here.

numba_dpex/spirv_generator.py
diptorupd: @mingjie-intel refer #798

diptorupd:
@oleksandr-pavlyk @mingjie-intel One side-effect of vendoring llvm-spirv is that we are now unable to build dpex properly using only `python setup.py develop`. I and Mingjie discussed a detour of adding a `build_locally.py` script that can build both llvm-spirv and dpex.

While that solves the issues, I still do not like the situation as some developer might prefer to use only setup.py to build dpex. Unlike dpctl, our solution actually makes that impossible. Before we merge I wanted to once more revisit some alternative options:

1.  Make `dpcpp_linux-x64` and `dpcpp_win-x64` packages a runtime dependency of `dpex` and make sure we use the` llvm-spirv` distributed in that package.
2.  Use the `llvm-spirv` conda package that Khronos group maintains. The previous llvm-spirv in IDP was a derivative of that and not well maintained by us. There is a possibility that some specific feature we need is available in dpcpp's version of llvm-spirv, but not yet in upstream.
3. Have a tools directory in dpex package to store the binary there and distribute it as a data file?

For any of these solutions to work, both conda and pip use cases need to be considered.

PR #803, opened 2022-10-18
Move target.py and descriptor.py into the core module.

diptorupd:
Moves the `numba_dpex.target` and `numba_dpex.descriptor` modules into `numba_dpex.core` sub-module and update all impacted imports.

The eventual goal of these incremental changes is to move all the compiler machinery into `core`.


diptorupd:
Merging based on public CI. Failures of teamcity CI is related to IPLDT check that is addressed in #802

PR #817, opened 2022-11-10
Removes the experimental parfor fallback feature.

diptorupd:
`numba_dpex` provides an experimental parfor fallback feature. Any parfor node that could not be lowered using dpex fell back to the default Numba parfor lower extension and was lowered as a parallel loop nest on the CPU.

The PR removes the fallback feature. Instead, dpex now will raise an exception informing users what code could not be lowered.

The benefit of the change is to clearly indicate to users when some parallel code could not be offloaded.

Supersedes #794 


diptorupd:
After #957 and all the dpnp parfor related changes, numpy offload is now deprecated and we will just remove these features in next release.

PR #839, opened 2022-12-07
Removes the numpy_usm_shared module from numba_dpex.

diptorupd:
The PR removes the `numpy_usm_shared` module from numba_dpex. All related unit test cases are also removed.


diptorupd:
accidentally closed.

PR #840, opened 2022-12-07
Update the usm_ndarray example.

diptorupd:
Fixes #839 

The PR updates the `usm_ndarray.py` to properly create `dpctl.tensor.usm_ndarray` arrays and pass them to a kernel instead of copying NumPy arrays to `dpctl.tensor.usm_ndarray`.


mingjie-intel (approved):
Tested locally, worked.

PR #833, opened 2022-12-03
A new LRU cache mechanism for numba_dpex

chudur-budur:
- [x] Have you provided a meaningful PR description?
    - This PR implements an LRU cache mechanism for numba_dpex that doesn't use file/pickle.
    - File/pickle saving capabilities might be added in the future (if at all necessary).
    - Extended WIP from https://github.com/IntelPython/numba-dpex/issues/814 and https://github.com/IntelPython/numba-dpex/pull/815
- [x] Have you added a test, reproducer or referred to an issue with a reproducer?
- [x] Have you tested your changes locally for CPU and GPU devices?
- [x] Have you made sure that new changes do not introduce compiler warnings?
- [x] If this PR is a work in progress, are you filing the PR as a draft?

PR #815, opened 2022-11-04
Using numba kernel caching for numba-dpex spirv kernel

chudur-budur:
- [x] Have you provided a meaningful PR description?
    - WIP for #814 
- [ ] Have you added a test, reproducer or referred to an issue with a reproducer?
- [x] Have you tested your changes locally for CPU and GPU devices?
- [ ] Have you made sure that new changes do not introduce compiler warnings?
- [x] If this PR is a work in progress, are you filing the PR as a draft?

Issue #814, opened 2022-11-01
Implementing a more robust caching mechanism


chudur-budur:
We need to have a more robust function/kernel caching mechanism. An example could be something like that has been implemented in `numba`.

https://numba.readthedocs.io/en/stable/developer/caching.html

https://github.com/numba/numba/blob/main/numba/core/caching.py

However the `numba` implementation uses files to save compiled (pickled) kernels. Here are some ideas:

 - Instead of using files, we can keep using the memory (dictionary)
 - Implement a queuing mechanism that prioritizes the compiled functions/kernels based on cache hit.
 - Optimize the file access: we will start saving in file only when the size of the dictionary reaches a certain limit, etc.

The caching will be used here:

https://github.com/IntelPython/numba-dpex/blob/main/numba_dpex/compiler.py

PR #843, opened 2022-12-08
A new LRU cache mechanism for numba_dpex with kernel pickling capability

chudur-budur:
    - This PR addresses issue #814 
    - This PR implements an LRU cache mechanism for numba_dpex.
    - The LRU cache can pickle kernels if they are evicted.
    - This PR is rebased on top of https://github.com/IntelPython/numba-dpex/commit/26c4334b1ef0b08ade41c5ca301c528ca2e9aa79
    - This PR supercedes #815 and #833

TODO:
- [x] Add cache capacity in config.
- [x] Remove OS-info from the key.
- [x] Split `driver.py` into separate commits.
- [x] Documentation about the caching mechanism. Add a new rst file in https://github.com/IntelPython/numba-dpex/tree/refactor/kernel_interfaces/docs/developer_guides
- [ ] Currently caching is initialized on kernel/function basis (i.e. local), we need to make it global (i.e. entire program basis).
- [x] Need to do similar caching for functions and function templates.

### DIFF
--- a/numba_dpex/core/caching.py
+++ b/numba_dpex/core/caching.py
@@ -1,0 +1,503 @@
+# SPDX-FileCopyrightText: 2020 - 2022 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import hashlib
+import sys
+from abc import ABCMeta, abstractmethod
+
+from numba.core.caching import CacheImpl, IndexDataCacheFile
+from numba.core.serialize import dumps
+
+from numba_dpex import config
+
+
+def build_key(sig, argtypes, pyfunc, codegen, backend=None, device_type=None):
+    """Constructs a key from python function, context, backend
+        and the device type.
+
+    Compute index key for the given signature and codegen.
+    It includes a description of the OS, target architecture
+    and hashes of the bytecode for the function and, if the
+    function has a __closure__, a hash of the cell_contents.
+
+    Args:
+        sig (inspect.Signature): The signature object of
+            a python function.
+        codegen (numba.core.codegen.Codegen):
+            The codegen object found from the target context.
+        backend (enum, optional): A 'backend_type' enum.
+            Defaults to None.
+        device_type (enum, optional): A 'device_type' enum.
+            Defaults to None.
+
+    Returns:
+        tuple: A tuple of signature, magic_tuple of codegen
+            and another tuple of hashcodes from bytecode and
+            cell_contents.
+    """
+
+    codebytes = pyfunc.__code__.co_code
+    if pyfunc.__closure__ is not None:
+        try:
+            cvars = tuple([x.cell_contents for x in pyfunc.__closure__])
+            # Note: cloudpickle serializes a function differently depending
+            #       on how the process is launched; e.g. multiprocessing.Process
+            cvarbytes = dumps(cvars)
+        except:
+            cvarbytes = b""  # a temporary solution for function template
+    else:
+        cvarbytes = b""
+
+    return (
+        sig,
+        argtypes,
+        codegen.magic_tuple(),
+        backend,
+        device_type,
+        (
+            hashlib.sha256(codebytes).hexdigest(),
+            hashlib.sha256(cvarbytes).hexdigest(),
+        ),
+    )
+
+
+class _CacheImpl(CacheImpl):
+    """Implementation of `CacheImpl` to be used by subclasses of `_Cache`.
+
+    This class is an implementation of `CacheImpl` to be used by subclasses
+    of `_Cache`. To be assigned in `_impl_class`. Implements the more common
+    and core mechanism for the caching.
+
+    """
+
+    def reduce(self, data):
+        """Serialize an object before caching.
+        Args:
+            data (object): The object to be serialized before pickling.
+        """
+        # TODO: Implement, but looks like we might not need it at all.
+        # Look at numba.core.caching for how to implement.
+        pass
+
+    def rebuild(self, target_context, reduced_data):
+        """Deserialize after unpickling from the cache.
+        Args:
+            target_context (numba_dpex.core.target.DpexTargetContext):
+                The target context for the kernel.
+            reduced_data (object): The data to be deserialzed after unpickling.
+        """
+        # TODO: Implement, but looks like we might not need it at all.
+        # Look at numba.core.caching for how to implement.
+        pass
+
+    def check_cachable(self, cres):
+        """Check if a certain object is cacheable.
+
+        Args:
+            cres (object): The object to be cached. For example, if the object is
+            `CompileResult`, then you might want to follow the similar checks as
+            has been done in `numba.core.caching.CompileResultCacheImpl`.
+
+        Returns:
+            bool: Return `True` if cacheable, otherwise `False`.
+        """
+        # TODO: Although, for the time being, assuming all Kernels in numba_dpex
+        # are always cachable. However, we might need to add some bells and
+        # whistles in the future. Look at numba.core.caching for how to implement.
+        return True
+
+
+class AbstractCache(metaclass=ABCMeta):
+    """Abstract cache class to specify basic caching operations.
+
+    This class will be used to create an non-functional dummy cache
+    (i.e. NullCache) and other functional cache. The dummy cache
+    will be used as a placeholder when caching is disabled.
+
+    Args:
+        metaclass (type, optional): Metaclass for the abstract class.
+            Defaults to ABCMeta.
+    """
+
+    @abstractmethod
+    def get(self):
+        """An abstract method to retrieve item from the cache."""
+
+    @abstractmethod
+    def put(self, key, value):
+        """An abstract method to save item into the cache.
+
+        Args:
+            key (object): The key for the data
+                (i.e. compiled kernel/function etc.).
+            value (object): The data (i.e. compiled kernel/function)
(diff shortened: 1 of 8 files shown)


numba_dpex/core/caching.py
```diff
@@ -0,0 +1,387 @@
+# Copyright 2022 Intel Corporation
```
diptorupd, line 1: Copyright header should follow style of rest of the files. We use `reuse` to generate the headers.

numba_dpex/core/caching.py
```diff
@@ -0,0 +1,387 @@
...
+from abc import ABCMeta, abstractmethod
+
+from numba.core.caching import CacheImpl, IndexDataCacheFile, _Cache
+from numba.core.serialize import dumps
+
+from numba_dpex import config
+
+
+def build_key(sig, pyfunc, codegen, backend=None, device_type=None):
+    """Constructs a key from the data object.
```
diptorupd, line 25: Comment needs clarity. Which argument is the "data" object?

numba_dpex/core/caching.py
```diff
@@ -0,0 +1,387 @@
...
+
+    Compute index key for the given signature and codegen. It includes
```
diptorupd, line 27: Does not seem right. Where are we passing OS information?

numba_dpex/core/caching.py
```diff
@@ -0,0 +1,387 @@
...
+            hashlib.sha256(codebytes).hexdigest(),
+            hashlib.sha256(cvarbytes).hexdigest(),
+        ),
+    )
+
+
+class _CacheImpl(CacheImpl):
+    """Implementation of `CacheImpl` to be used by subclasses of `_Cache`.
+
+    This class is an implementation of `CacheImpl` to be used by subclasses of `_Cache`.
```
diptorupd, line 67: Check for 80 char line width of comments.

numba_dpex/core/caching.py
```diff
@@ -0,0 +1,387 @@
...
+
+    """
+
+    def reduce(self, data):
+        """Serialize an object before caching.
+
+        Args:
+            data (object): The object to be serialized before pickling.
+        """
+        # TODO: Implement, but looks like we might not need it at all.
```
diptorupd, line 79: If the function is not needed, I prefer we remove it.

numba_dpex/core/caching.py
```diff
@@ -0,0 +1,387 @@
...
+        # Look at numba.core.caching for how to implement.
+        pass
+
+    def rebuild(self, target_context, reduced_data):
+        """Deserialize after unpickling from the cache.
```
diptorupd, line 84: same here.

numba_dpex/core/caching.py
```diff
@@ -0,0 +1,387 @@
...
+        # TODO: Although, for the time being, assuming all Kernels in numba_dpex
+        # are always cachable. However, we might need to add some bells and
+        # whistles in the future. Look at numba.core.caching for how to implement.
+        return True
+
+
+class AbstractCache(metaclass=ABCMeta):
+    @abstractmethod
+    def get(self):
+        """_summary_"""
```
diptorupd, line 115: Docstring is a stub

numba_dpex/core/caching.py
```diff
@@ -0,0 +1,387 @@
...
+        Returns:
+            str: str: A human readable representation of a Node.
+        """
+        return "(" + str(self.key) + ": " + str(self.value) + ")"
+
+    def __repr__(self):
+        """__repr__ for Node
+
+        Returns:
+            sstr: A human readable representation of a Node.
```
diptorupd, line 165: Return type is `sstr` or is it a typo?

numba_dpex/core/caching.py
```diff
@@ -0,0 +1,387 @@
...
+
+
+class LRUCache(AbstractCache):
+    """LRUCache implementation for caching kernels, functions and modules
+
+    The cache is basically a doubly-linked-list backed with a dictionary
+    as a lookup table.
+    """
+
+    def __init__(self, capacity=10, pyfunc=None):
```
diptorupd, line 177: As we were discussing, preferably we should define the cache limit in MB for the size of cached objects, rather than the number of cached objects.

numba_dpex/core/caching.py
```diff
@@ -0,0 +1,387 @@
...
+
+        Returns:
+            str: A human readable representation of the cache.
+        """
+        return self.__str__()
+
+    def clean(self):
+        """Clean the cache"""
+        self._lookup = {}
+        # self._evicted = {}
```
diptorupd, line 259: Remove commented out line.

numba_dpex/core/caching.py
```diff
@@ -0,0 +1,387 @@
...
+
+        Args:
+            capacity (int, optional): The max capacity of the cache. Defaults to 10.
+            pyfunc (NoneType, optional): A python function to be cached. Defaults to None.
+        """
+        self._capacity = capacity
+        self._lookup = {}
+        self._evicted = {}
+        self._dummy = Node(0, 0)
+        self._head = self._dummy.next
```
diptorupd, line 188: Where are we setting `_dummy.next`? The `Node` class does not have a `next` attribute.

numba_dpex/core/kernel_interface/dispatcher.py
```diff
@@ -57,6 +59,7 @@ def __init__(
         debug_flags=None,
         compile_flags=None,
         array_access_specifiers=None,
+        enable_cache=True,
```
diptorupd, line 62: Why are we updating this function? The function is a way to keep track of an optional "array access specifier" (readonly, write, read-write) for a NumPy array passed to a kernel as an argument. 

The `get_ordered_arg_access_types` function should be unrelated to the kernel caching mechanism. Moreover, I am going to remove the function once we remove NumPy argument support for kernels.

numba_dpex/core/kernel_interface/dispatcher.py
```diff
@@ -420,24 +446,51 @@ def __call__(self, *args, global_range=None, local_range=None):
...
 
-        # TODO: Enable caching of kernels, but do it using Numba's caching
-        # machinery
-
-        kernel = SpirvKernel(self.pyfunc, self.kernel_name)
-        kernel.compile(
-            arg_types=argtypes,
-            debug=self.debug_flags,
-            extra_compile_flags=self.compile_flags,
+        # TODO: Enable caching of kernels, but do it using LRU
```
diptorupd, line 449: Can be removed?

diptorupd (changes requested):
@chudur-budur I left few nits about comments and such.

The main design issue I have is that the cache is per kernel/dispatcher. The LRU cache should be a global object that dpex stores for all kernels.

numba_dpex/core/caching.py
chudur-budur: How do I use `reuse` to generate a correct header? I couldn't find any pointers in the repo.

chudur-budur: > Copyright header should follow style of rest of the files. We use `reuse` to generate the headers.

How do I use `reuse` to generate a correct header? I couldn't find any pointers in the repo.

numba_dpex/core/caching.py
chudur-budur: We are building a key with backend and device_type. This function is called in dispatcher (before caching/decaching). But where do I get OS info?

numba_dpex/core/caching.py
chudur-budur: The `Node` class has `next` and `previous`.

numba_dpex/core/kernel_interface/dispatcher.py
chudur-budur: Yes, because this part has been executed in the next lines, please follow the `if-else` block after that.

numba_dpex/core/caching.py
chudur-budur: > As we were discussing, preferably we should define the cache limit in MB for the size of cached objects, rather than the number of cached objects.

That approach will have some problems. For example, let's say at a certain state of the runtime `t1`, we have an LRU cache of kernels as follows:

```
cache(t1) = [5kb, 5kb, 5kb, 5kb]
```

Let's say the cache size limit is 20Kb. Now assume a new kernel of size 10kb wants to be cached. In that case we need to evict 2 kernels to make room for the new one. 

However, in the current implementation, we can evict just one kernel to make room for one new kernel.

chudur-budur:
> @chudur-budur I left few nits about comments and such.
> 
> The main design issue I have is that the cache is per kernel/dispatcher. The LRU cache should be a global object that dpex stores for all kernels.

Yea we can easily move the cache into global. We need to discuss about the strategy. Btw, numba does caching per-function/kernel basis. I am not quite sure why that's the case.

numba_dpex/core/caching.py
diptorupd: Ok. Let us keep it as is for now.

numba_dpex/core/caching.py
chudur-budur: We do not need this.

chudur-budur:
1. Fixed failed test cases.
2. There were couple of bugs in the code, fixed all of them.
3. Now it should work in all situations.

diptorupd (approved):

PR #804, opened 2022-10-18
Refactor/kernel interfaces

diptorupd:
- [x] Have you provided a meaningful PR description?

  - [x] The concept of a kernel was decoupled from the notion of dispatching a kernel to a device. The old implementation in `numba_dpex\compiler.py` mixes both things, making hard the separation of compute-follows-data based kernel launch from legacy `dpctl.device_context` based behavior. 
  - [x] Deprecates support for numpy arrays as kernel args.
  - [x] Deprecates support for `dpctl.device_context` for kernels.
  - [ ] Deprecates support for the square bracket notation using `__getitem__` to provide global and local ranges for a kernel launch. (to be reevaluated #790)
  - [x] Changes the behavior of specializing a kernel using only a signature. The new way to specialize will require a device type and a backend.
  - [x] Add validations for ndrange kernel launch parameters.
  - [ ] #862
  - [x] Improvements to exception messages using custom exceptions.
  - [x] The new API is now inside `numba_dpex.core.kernel_interface`.
  - [x] Migrate `DpexFunc` to new API
  - [x] Migrate parfor compilation to new API
  - [x] Remove `numba_dpex.compiler.py`
  - [x] Adds new unit test cases
  - [x] Enable caching of kernels in new API
  - [x] Doc strings
  - [x] Update examples
  - [ ] Generate API docs

- [X] Have you added a test, reproducer or referred to an issue with a reproducer?
- [x] Have you tested your changes locally for CPU and GPU devices?
- [x] Have you made sure that new changes do not introduce compiler warnings?
- [ ] If this PR is a work in progress, are you filing the PR as a draft?

Fixes #814, #816, #780, #810

PR #859, opened 2023-01-09
Added checks for setting up ndrange 

mingjie-intel:
Tow checks added for ndrange:

1. global_range and local_range have the same dimension. 
2. global_range must can divide local_range evenly. 
3. local_range should be within device allowed work item limit.


diptorupd:
@mingjie-intel I cherry-picked your commits into #804. I am closing the PR.

PR #866, opened 2023-01-11
Remove NumPy argument support for dpex kernels

mingjie-intel:
We will stop supporting NumPy arrays as arguments  to kernel. The following API changes need to be made:

- [x] Remove array_access_specifiers from JitKernel init
- [x] Remove get_ordered_arg_access_types
- [x] Simplify JitKernel._determine_kernel_launch_queue
- [x] Remove the special case for calling npytypes_array_to_dpex_array inside `target.TypingContext`.  Instead if a NumPy array argument is detected raise an exception.
- [x] Remove access types from Packer
- [x] Remove support for array from Packer._unpack_arguments
- [x] Remove _pack_array from Packer and sycl_queue in __init__


diptorupd:
Superseded by #1049

PR #878, opened 2023-01-19
Pass dpnp arrays directly to kernel instead of using dpctl.asarray.

diptorupd:
The changes in #804 ensure that dpnp arrays can be directly passed to a kernel as arguments without needing `asarray` conversion. Other updates to the test case to bring it in line with changes post #804.


diptorupd:
Closed in favour of #880

Issue #868, opened 2023-01-11
Conflicting SPIR-V versions when linking to atomic-ops.spir


fcharras:
I'm having troubles with using atomics from `atomic-ops.spir`, with the following error message:

```
...<elided traceback>...

popenargs = (['spirv-link', '--allow-partial-linkage', '-o', '/tmp/tmpcp_knjn_/2-linked-spirv', '/tmp/tmpcp_knjn_/1-generated-spirv', '/opt/venv/lib/python3.9/site-packages/numba_dpex/ocl/atomics/atomic_ops.spir'],)
kwargs = {}, retcode = 1
cmd = ['spirv-link', '--allow-partial-linkage', '-o', '/tmp/tmpcp_knjn_/2-linked-spirv', '/tmp/tmpcp_knjn_/1-generated-spirv', '/opt/venv/lib/python3.9/site-packages/numba_dpex/ocl/atomics/atomic_ops.spir']

    def check_call(*popenargs, **kwargs):
        """Run command with arguments.  Wait for command to complete.  If
        the exit code was zero then return, otherwise raise
        CalledProcessError.  The CalledProcessError object will have the
        return code in the returncode attribute.
    
        The arguments are the same as for the call function.  Example:
    
        check_call(["ls", "-l"])
        """
        retcode = call(*popenargs, **kwargs)
        if retcode:
            cmd = kwargs.get("args")
            if cmd is None:
                cmd = popenargs[0]
>           raise CalledProcessError(retcode, cmd)
E           subprocess.CalledProcessError: Command '['spirv-link', '--allow-partial-linkage', '-o', '/tmp/tmpcp_knjn_/2-linked-spirv', '/tmp/tmpcp_knjn_/1-generated-spirv', '/opt/venv/lib/python3.9/site-packages/numba_dpex/ocl/atomics/atomic_ops.spir']' returned non-zero exit status 1.

/opt/pyenv/versions/3.9.16/lib/python3.9/subprocess.py:373: CalledProcessError

error: 1: Conflicting SPIR-V versions: 1.4 (input modules 1 through 1) vs 1.0 (input module 2).

Traceback (most recent call last):
  File "/opt/venv/lib/python3.9/site-packages/numba_dpex/spirv_generator.py", line 137, in __del__
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpcp_knjn_/2-linked-spirv'
```

[...]

PR #880, opened 2023-01-19
Fix dpnp type

mingjie-intel:
1. Moved dpnp types.py typesof.py into core.
2. Let dpnp ndarray inherite from USM ndarray. 

Fixes #857

PR #885, opened 2023-01-22
Set the max spirv version to 1.1

diptorupd:
numba-dpex does not pin the max-spirv-version that `llvm-spirv` generates and lets `llvm-spriv` decide what SPIR-V code to generate based on the LLVM-IR generated by the Numba front-end.

The behaviour is causing an issue with oneAPI dpcpp 2023.0 (Refer #868). The issue arises when a pre-compiled SPIR-V binary produced by compiling an OpenCL program implementing software atomic add for floating point values is linked to the SPIR-V module generated by JIT compilation inside numba-dpex.

The PR implements a workaround by setting the `spirv-max-version` flags to 1.1 when generating SPIR-V both for the pre-compiled OpenCL program and inside numba-dpex. SPIR-V version 1.1 is used as it seems to be the default used by dpcpp.

### DIFF
--- a/numba_dpex/spirv_generator.py
+++ b/numba_dpex/spirv_generator.py
@@ -75,6 +75,8 @@ class CmdLine:
         if config.DEBUG:
             llvm_spirv_flags.append("--spirv-debug-info-version=ocl-100")
 
+        if not config.NATIVE_FP_ATOMICS:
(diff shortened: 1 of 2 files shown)


diptorupd:
Merging as the commit has already been tested in #880

Issue #886, opened 2023-01-23
Perormance regressions introduced by latest changes to main


diptorupd:
> I also see segfaults, makes me think that something is wrong with local memory.

_Originally posted by @fcharras in https://github.com/IntelPython/numba-dpex/issues/816#issuecomment-1400514147_

> The new problems are not related to #877 but to #876 , reported in https://github.com/IntelPython/numba-dpex/issues/816

_Originally posted by @fcharras in https://github.com/IntelPython/numba-dpex/issues/867#issuecomment-1400406083_

fcharras:
@diptorupd @chudur-budur

We're having conversations about this in multiple separate issues, let's keep it in this dedicated thread ?

> > I also see segfaults, makes me think that something is wrong with local memory.
> 
> The problem most likely is because I without realizing the impact changed the way the global and local range args passed to the `__getitem__`. We are working on reverting my changes and fixing the issue.

Do you still need a minimal reproducer ? I haven't come up with one yet. Yet again it's not trivial to reproduce, simple kernels still work.

If you have a commit that revert those changes I can easily say if it fixes the issues though. (basically running the benchmark script in https://github.com/soda-inria/sklearn-numba-dpex )

diptorupd:
@fcharras No we do not need th reproducer. Let @chudur-budur open the PR and then you can test and let us know if the issue is fixed.

chudur-budur:
@fcharras Could you please test PR #888 and check if you are seeing any performance issues?

fcharras:
Unfortunately, issues persists in #888 :-(

I just realized that I made a mistake in my bug report that might have misled you. The problems start at the merge commit of https://github.com/IntelPython/numba-dpex/pull/804 not https://github.com/IntelPython/numba-dpex/pull/876 . The merge commits of those two PRs are very close to each other and I missclicked the copy/paste :-/

chudur-budur:
@fcharras Could please test #896 and see if you are getting any slowdown?

fcharras:
@chudur-budur @diptorupd please see https://github.com/IntelPython/numba-dpex/issues/898 that give a minimal reproducer for what caused the regression I was facing, which is more than a performance issue.

[...]

Issue #871, opened 2023-01-18
2D vector sum using usm_ndarrays does not work in kernels


diptorupd:
Two dimensional indexing over `dpctl.tensor.usm_ndarray` data types inside a kernel does not work. Incorrect result is generated.

Reproducer:

```
import dpctl
import dpctl.tensor as dpt
import numba_dpex as dpex

@dpex.kernel
def data_parallel_sum(a, b, c):
    """
    A two-dimensional vector addition example using the ``kernel`` decorator.
    """
    i = dpex.get_global_id(0)
    j = dpex.get_global_id(1)
    c[i, j] = a[i, j] + b[i, j]

# Array dimensions
X = 8
Y = 8
global_size = X, Y

device = dpctl.select_default_device()
a_dpt = dpt.arange(X * Y, dtype=dpt.float32, device=device)
a_dpt = dpt.reshape(a_dpt, (X, Y))
b_dpt = dpt.arange(X * Y, dtype=dpt.float32, device=device)
b_dpt = dpt.reshape(b_dpt, (X, Y))
c_dpt = dpt.empty_like(a_dpt)
c_dpt = dpt.reshape(c_dpt, (X, Y))

data_parallel_sum[global_size](a, b, c)

```

diptorupd:
A cleaned up LLVM IR for the kernel is as follows. As can be seen the indexing is being calculated incorrectly.

For shape (X,Y),

global id (0,0) : 0 (OK serendipitous) 
global id (0,1) : Y (should be 1)
global id (0,2) : 2+2Y (should be 2)

```
; ModuleID = '/home/diptorupd/Desktop/vector_sum2d.ll'
source_filename = "/home/diptorupd/Desktop/vector_sum2d.ll"
target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
target triple = "spir64-unknown-unknown"

@_data_parallel_sum_ = common local_unnamed_addr global i8* null

; Function Attrs: alwaysinline
define spir_func i32 @_data_parallel_sum_(
    i8** nocapture writeonly %.ret, 
    float addrspace(1)* nocapture readnone %arg.a.0, 
    float addrspace(1)* nocapture readnone %arg.a.1, 
    i64 %arg.a.2, 
    i64 %arg.a.3, 
    float addrspace(1)* nocapture readonly %arg.a.4, 
    i64 %arg.a.5.0, 
[...]

Issue #867, opened 2023-01-11
Can't reuse `dpex.func` device functions  with different signatures


fcharras:
This used to work with `numba_dpex<=0.18.1` but fails with `>=0.19`:

```python
import numba_dpex as dpex
import dpctl.tensor as dpt
import numpy as np


@dpex.func
def g(array_in, idx, const):
    array_in[idx] = const

@dpex.kernel
def kernel_a(array_in):
    idx = dpex.get_global_id(0)
    g(array_in, idx, np.int64(0))

@dpex.kernel
def kernel_b(array_in):
    idx = dpex.get_global_id(0)
    g(array_in, idx, np.int32(0))  # NB: call with inputs of different types than in kernel_a

   
dtype = np.float32
size = 16
array_in = dpt.zeros(sh=(size,), dtype=dtype)

kernel_a[size, size](array_in)
kernel_b[size, size](array_in)
```

with `numba_dpex>=0.19` this snippet gives the following exception:

```
<...traceback elided...>
LoweringError: Failed in dpex_nopython mode pipeline (step: Custom Lowerer with auto-offload support)
No definition for lowering <function get_global_id at 0x7f47c2edf700>(uint32,) -> int64

File "<ipython-input-14-e32b2bd5f6a0>", line 17:
def kernel_b(array_in):
    idx = dpex.get_global_id(0)
    ^

During: lowering "idx = call $4load_method.1($const6.2, func=$4load_method.1, args=[Var($const6.2, <ipython-input-14-e32b2bd5f6a0>:17)], kws=(), vararg=None, varkwarg=None, target=None)" at <ipython-input-14-e32b2bd5f6a0> (17)
```

There's a workaround that consists in ensuring that in both kernels the device function is at different memory locations with a different name, e.g.:

```python
import numba_dpex as dpex
import dpctl.tensor as dpt
import numpy as np


def make_g():
    @dpex.func
    def g(array_in, idx, const):
        array_in[idx] = const
    return g

g = make_g()
@dpex.kernel
def kernel_a(array_in):
    idx = dpex.get_global_id(0)
    g(array_in, idx, np.int64(0))

g_ = make_g()
@dpex.kernel
def kernel_b(array_in):
    idx = dpex.get_global_id(0)
    g_(array_in, idx, np.int32(0))

   
dtype = np.float32
size = 16
array_in = dpt.zeros(sh=(size,), dtype=dtype)

kernel_a[size, size](array_in)
[...]

Issue #790, opened 2022-10-07
Deprecate __getitem__ support in numba_dpex.kernel


diptorupd:
The numba_dpex.kernel decorator currently supports passing the `global_size` and `local_size` parameters while launching a kernel using the indexing operator (`[] operator`). Internally, it is implemented by overriding the `__getitem__` function in the `numba_dpex.compiler.KernelBase` class.

The design is a legacy of how `numba.cuda` does it and we are mimicking their design. 

However, IMO there are several problems with the design. 

- Programmers need to be careful that they enter the parameters in the right order.
- The usage is not clean as it does not allow specifying the arguments as keywords.
- It is not clear to users that the first argument that corresponds to the global range is required while the second argument corresponding to the local range is optional.

I am in favour of deprecating the feature in 0.19 (next release) and instead asking users to pass the global range and local range using keyword arguments. Of these the global range will be a required argument, but the local range will be optional.

Creating the issue to start a design discussion.

diptorupd:
ping @fcharras @ogrisel @adarshyoga @oleksandr-pavlyk @mingjie-intel @chudur-budur

diptorupd:
Would not change. Based on discussions the `__getitem__` has been retained but going forward will require `Range` or `NdRange` args.

PR #888, opened 2023-01-25
A more consistent kernel launch parameter syntax

chudur-budur:
Addressing issues #867 and #871. This might also fix issue #886 
Also related to issues #790 and #766


numba_dpex/core/kernel_interface/dispatcher.py
```diff
@@ -468,52 +522,56 @@ def __getitem__(self, args):
...
-                elif len(gr) != 0 and all(isinstance(v, int) for v in gr):
-                    self._global_range = list(gr)
-                else:
-                    raise IllegalRangeValueError(kernel_name=self.kernel_name)
 
-                if isinstance(lr, int):
-                    self._local_range = [lr]
-                elif isinstance(lr, list) and len(lr) == 0:
-                    # deprecation warning
+        if isinstance(args, Ranges):
```
diptorupd, line 526: Instead of a single class, my preference is to have two classes `Range` and NdRange`. I think the `Ranges` class can be renamed to `NdRange` with the `local_range` arg as required. A similar class with only one parameter will represent `Range`.

numba_dpex/core/kernel_interface/dispatcher.py
```diff
@@ -468,52 +522,56 @@ def __getitem__(self, args):
...
+                isinstance(args, tuple)
+                and len(args) == 2
+                and isinstance(args[0], int)
+                and isinstance(args[1], int)
+            ):
+                warn(
+                    "Ambiguous kernel launch paramters. "
+                    + "If your data have dimensions > 1, "
+                    + "include a default/empty local_range. "
+                    + "i.e. <function>[(M,N), numba_dpex.DEFAULT_LOCAL_RANGE](<params>), "
```
diptorupd, line 540: The message is OK in it self. But, if we intend to totally deprecate passing tupples or lists to `__getitem__` let us tell the users about the new Range and NdRange classes.

numba_dpex/core/kernel_interface/dispatcher.py
```diff
@@ -468,52 +522,56 @@ def __getitem__(self, args):
...
 
-                if isinstance(lr, int):
-                    self._local_range = [lr]
-                elif isinstance(lr, list) and len(lr) == 0:
-                    # deprecation warning
+        if isinstance(args, Ranges):
+            self._global_range = list(args.global_range)[::-1]
+            self._local_range = list(args.local_range)[::-1]
+        else:
+            if (
```
diptorupd, line 530: Print a deprecation notice right here asking users to start using `Range` and `NdRange` classes.

numba_dpex/core/kernel_interface/dispatcher.py
```diff
@@ -441,6 +445,56 @@ def _determine_kernel_launch_queue(self, args, argtypes):
             else:
                 raise ExecutionQueueInferenceError(self.kernel_name)
 
+    def _raise_invalid_kernel_enqueue_args(self):
+        error_message = (
+            "Incorrect number of arguments for enqueuing numba_dpex.kernel. "
+            "Usage: device_env, global size, local size. "
```
diptorupd, line 451: The error message is wrong and seems to have been not updated for the last two years even as we made many API changes!! Please do not use this old error message.

numba_dpex/core/kernel_interface/dispatcher.py
```diff
@@ -468,52 +522,56 @@ def __getitem__(self, args):
...
+                return self
+
+            if not isinstance(args, Iterable):
+                args = [args]
+
+            ls = None
+            nargs = len(args)
+            # Check if the kernel enquing arguments are sane
+            if nargs < 1 or nargs > 2:
+                self._raise_invalid_kernel_enqueue_args()
```
diptorupd, line 555: Directly raise `InvalidKernelLaunchArgsError` here the old `_raise_invalid_kernel_enqueue_args` has a totally wrong error message.

numba_dpex/core/kernel_interface/dispatcher.py
```diff
@@ -468,52 +522,56 @@ def __getitem__(self, args):
...
+
+            gs = self._ensure_valid_work_item_grid(args[0])
+            # If the optional local size argument is provided
+            if nargs == 2:
+                if args[1] != []:
+                    ls = self._ensure_valid_work_group_size(args[1], gs)
```
diptorupd, line 561: No need to call `_ensure_valid_work_group_size`. The function can be removed altogether. We care checking ranges for validity in the new `_check_ranges` function.

numba_dpex/examples/kernel/vector_sum2D.py
```diff
@@ -22,7 +22,7 @@ def data_parallel_sum(a, b, c):
 
 
 def driver(a, b, c, global_size):
-    data_parallel_sum[global_size](a, b, c)
+    data_parallel_sum[global_size, dpex.DEFAULT_LOCAL_SIZE](a, b, c)
```
diptorupd, line 25: Ok. But, how about demonstrating the use of the new and likely future method of using Range classes?

numba_dpex/core/kernel_interface/dispatcher.py
chudur-budur: Done.

numba_dpex/core/kernel_interface/dispatcher.py
chudur-budur: Done.

numba_dpex/core/kernel_interface/dispatcher.py
chudur-budur: Done

numba_dpex/core/kernel_interface/dispatcher.py
chudur-budur: Fixed.

numba_dpex/core/kernel_interface/dispatcher.py
chudur-budur: Old functions are removed.

numba_dpex/core/kernel_interface/dispatcher.py
chudur-budur: Old functions are removed.

numba_dpex/examples/kernel/vector_sum2D.py
chudur-budur: A new driver function is added to showcase the new way of specifying kernel lauch params.

numba_dpex/core/kernel_interface/dispatcher.py
```diff
@@ -468,51 +470,79 @@ def __getitem__(self, args):
...
-                if isinstance(gr, int):
-                    self._global_range = [gr]
-                elif len(gr) != 0 and all(isinstance(v, int) for v in gr):
-                    self._global_range = list(gr)
-                else:
-                    raise IllegalRangeValueError(kernel_name=self.kernel_name)
+        warn(
+            """
+            The current syntax for specification of kernel lauch parameters
+            will be deprecated soon. Users are requested to do it through
```
diptorupd, line 476:
```suggestion
            is deprecated. Users should set the kernel launch arguments through
```

diptorupd: Moreover, why should I get the deprecation warning if I am using the right API. The warning should be moved into the else block.

numba_dpex/core/kernel_interface/dispatcher.py
```diff
@@ -468,51 +470,79 @@ def __getitem__(self, args):
...
+            Range/NdRange classes.
+                Example:
+                    from numba_dpex.core.kernel_interface.utils import NdRange
+                    ranges = NdRange((8,8), (2,2)) # 2D global and local ranges
+                    <function>[ranges](<parameters>)
+            """,
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        if isinstance(args, NdRange):
```
diptorupd, line 486: The change does not match SYCL semantics. The local range for an NdRange is not optional. We really need two classes `Range` and `NdRange`. There is a difference in how a kernel is submitted based on  whether a range or an ndrange is provided. 

We have to be clear and explicit about the usage. When using NdRange class it should without ambiguity indicate that the kernel will be submitted as an ndrange kernel.

chudur-budur: @diptorupd Right now, the correct syntax for kernel parameters are like one of these:

```python
ranges = NdRange((2,2,2),(1,1,1))
func[ranges](a,b,c)
```

or 

```python
ranges = NdRange((2,2,2))
func[ranges](a,b,c)
```

or

```python
global_range = Range(2,2,2)
local_range = Range(1,1,1)
ranges = NdRange(global_range, local_range)
func[ranges](a,b,c)
```

or

```python
global_range = Range(2,2,2)
ranges = NdRange(global_range)
func[ranges](a,b,c)
```

If these are not correct, could you give me an example of a correct syntax?

numba_dpex/core/kernel_interface/dispatcher.py
chudur-budur: Fixed.

numba_dpex/core/kernel_interface/dispatcher.py
diptorupd: In the above examples, (1) and (3) is what we should be support.
(2) and (4) we should not.

That is, `NdRange` should require both `global_range` and `local_range` to be set. We should not allow `NdRange` where `local_range` is `None`.

chudur-budur: > We should not allow `NdRange` where `local_range` is `None`.

But there are many examples where we don't know the `local_range` and therefore we use `numba_dpex.DEFAULT_LOCAL_SIZE` (which is basically a `None`). For example, what do we do in the `sum2D.py` in the `examples` folder?

chudur-budur: Fixed.

numba_dpex/config.py
```diff
@@ -99,7 +99,7 @@ def __getattr__(name):
 ENABLE_CACHE = _readenv("NUMBA_DPEX_ENABLE_CACHE", int, 1)
 # Capacity of the cache, execute it like:
 #   NUMBA_DPEX_CACHE_SIZE=20 python <code>
-CACHE_SIZE = _readenv("NUMBA_DPEX_CACHE_SIZE", int, 10)
+CACHE_SIZE = _readenv("NUMBA_DPEX_CACHE_SIZE", int, 128)
```
diptorupd, line 102: Why the change?

diptorupd (approved):
LGTM! Thank you @chudur-budur

PR #895, opened 2023-02-01
Update gitignore to ignore vscode settings files.

diptorupd:
Add vscode settings files to gitignore


mingjie-intel (approved):
Good to go.

PR #896, opened 2023-02-02
Solving perferomance regression issue by caching the kernel_bundle

chudur-budur:
This PR might solve issue #886. 

In version 0.19.0, the `create_program_from_spirv()` was done only once. See [line 483](https://github.com/IntelPython/numba-dpex/blob/0.19.0/numba_dpex/compiler.py#L483) in `compiler.py`.

In the current main it's been done everytime when `__call__()` function was invoked from `JitKernel` (See [here](https://github.com/IntelPython/numba-dpex/blob/main/numba_dpex/core/kernel_interface/dispatcher.py#L630)). Therefore we are now caching the `kernel_bundle` to avoid the repeated call of `create_program_from_spirv()`. 

### DIFF
--- a/numba_dpex/core/caching.py
+++ b/numba_dpex/core/caching.py
@@ -16,8 +16,8 @@ from numba_dpex.core.types import USMNdArray
 def build_key(
     argtypes, pyfunc, codegen, backend=None, device_type=None, exec_queue=None
 ):
-    """Constructs a key from python function, context, backend and the device
-    type.
+    """Constructs a key from python function, context, backend, the device
+    type and execution queue.
 
     Compute index key for the given argument types and codegen. It includes a
     description of the OS, target architecture and hashes of the bytecode for
@@ -34,6 +34,7 @@ def build_key(
             Defaults to None.
         device_type (enum, optional): A 'device_type' enum.
             Defaults to None.
+        exec_queue (dpctl._sycl_queue.SyclQueue', optional): A SYCL queue object.
 
     Returns:
         tuple: A tuple of return type, argtpes, magic_tuple of codegen
--- a/numba_dpex/core/caching.py
+++ b/numba_dpex/core/caching.py
@@ -13,7 +13,9 @@ from numba_dpex import config
 from numba_dpex.core.types import USMNdArray
 
 
-def build_key(argtypes, pyfunc, codegen, backend=None, device_type=None):
+def build_key(
+    argtypes, pyfunc, codegen, backend=None, device_type=None, exec_queue=None
+):
     """Constructs a key from python function, context, backend and the device
     type.
 
@@ -64,6 +66,7 @@ def build_key(argtypes, pyfunc, codegen, backend=None, device_type=None):
         codegen.magic_tuple(),
         backend,
         device_type,
+        exec_queue,
         (
             hashlib.sha256(codebytes).hexdigest(),
             hashlib.sha256(cvarbytes).hexdigest(),
--- a/numba_dpex/core/kernel_interface/dispatcher.py
+++ b/numba_dpex/core/kernel_interface/dispatcher.py
@@ -88,14 +88,21 @@ class JitKernel:
         # caching related attributes
         if not config.ENABLE_CACHE:
             self._cache = NullCache()
+            self._kernel_bundle_cache = NullCache()
         elif enable_cache:
             self._cache = LRUCache(
                 name="SPIRVKernelCache",
                 capacity=config.CACHE_SIZE,
                 pyfunc=self.pyfunc,
             )
+            self._kernel_bundle_cache = LRUCache(
+                name="KernelBundleCache",
+                capacity=config.CACHE_SIZE,
+                pyfunc=self.pyfunc,
+            )
         else:
             self._cache = NullCache()
+            self._kernel_bundle_cache = NullCache()
         self._cache_hits = 0
 
         if array_access_specifiers:
@@ -627,12 +634,26 @@ class JitKernel:
                     cache=self._cache,
                 )
 
-        # create a sycl::KernelBundle
-        kernel_bundle = dpctl_prog.create_program_from_spirv(
-            exec_queue,
-            device_driver_ir_module,
-            " ".join(self._create_sycl_kernel_bundle_flags),
+        kernel_bundle_key = build_key(
+            tuple(argtypes),
+            self.pyfunc,
+            dpex_kernel_target.target_context.codegen(),
+            exec_queue=exec_queue,
         )
+
+        artifact = self._kernel_bundle_cache.get(kernel_bundle_key)
+
+        if artifact is None:
+            # create a sycl::KernelBundle
+            kernel_bundle = dpctl_prog.create_program_from_spirv(
+                exec_queue,
+                device_driver_ir_module,
+                " ".join(self._create_sycl_kernel_bundle_flags),
+            )
+            self._kernel_bundle_cache.put(kernel_bundle_key, kernel_bundle)
+        else:
+            kernel_bundle = artifact
+
         #  get the sycl::kernel
         sycl_kernel = kernel_bundle.get_sycl_kernel(kernel_module_name)
 
--- a/numba_dpex/tests/kernel_tests/test_arg_types.py
+++ b/numba_dpex/tests/kernel_tests/test_arg_types.py
@@ -2,12 +2,14 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import sys
+
 import dpctl
+import dpctl.tensor as dpt
 import numpy as np
 import pytest
 
 import numba_dpex as dpex
-from numba_dpex.tests._helper import filter_strings
 
 global_size = 1054
 local_size = 1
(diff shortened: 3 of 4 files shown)


chudur-budur:
@adarshyoga Could you please test this PR to see if you are getting any slow down?

adarshyoga:
Are you caching both llvm bit code and sycl program generated from spirv?

.pre-commit-config.yaml
```diff
@@ -23,7 +23,7 @@ repos:
     -   id: blacken-docs
         additional_dependencies: [black==22.10]
 -   repo: https://github.com/pycqa/isort
-    rev: 5.10.1
+    rev: 5.12.0
```
mingjie-intel, line 26: Anything thing wrong with 10.1 or is there some feature we need for 12.0?

mingjie-intel (commented):
Please let Adarsh test before merge. thanks.

.pre-commit-config.yaml
chudur-budur: pre-commit fails to install isort 5.10.1 for [this reason](https://stackoverflow.com/questions/75269700/pre-commit-fails-to-install-isort-5-11-4-with-error-runtimeerror-the-poetry-co), but it can install isort 5.12.0 as suggested in the answer.

chudur-budur:
> Are you caching both llvm bit code and sycl program generated from spirv?

yes @adarshyoga

numba_dpex/core/kernel_interface/dispatcher.py
```diff
@@ -95,6 +95,11 @@ def __init__(
...
                 pyfunc=self.pyfunc,
             )
+            self._kernel_bundle_cache = LRUCache(
+                name="KernelBundleCache",
+                capacity=config.CACHE_SIZE,
+                pyfunc=self.pyfunc,
+            )
         else:
             self._cache = NullCache()
         self._cache_hits = 0
```
adarshyoga, line 105: Is _cache_hits being used? if this is only for book-keeping during debugging then it needs to be removed.

adarshyoga:
The performance regression in blackscholes no longer happens with this fix. See below times from blackscholes which are close to what we were getting with 0.19.0.

-------
ERF: Numba@jit-loop-par | Size:     524288 | MOPS:    490767591.93 | TIME:   0.002137
Time of first run  0.0021376459999373765
ERF: Numba@jit-loop-par | Size:    1048576 | MOPS:   1192850017.17 | TIME:   0.001758
Time of first run  0.0022624360008194344
ERF: Numba@jit-loop-par | Size:    2097152 | MOPS:   2397245139.53 | TIME:   0.001750
Time of first run  0.002746018000834738
ERF: Numba@jit-loop-par | Size:    4194304 | MOPS:   4262913460.23 | TIME:   0.001968
Time of first run  0.003016001999640139
ERF: Numba@jit-loop-par | Size:    8388608 | MOPS:   6976956836.03 | TIME:   0.002405
Time of first run  0.003975543000706239
ERF: Numba@jit-loop-par | Size:   16777216 | MOPS:  10806267130.05 | TIME:   0.003105
Time of first run  0.006068818000130705
ERF: Numba@jit-loop-par | Size:   33554432 | MOPS:  14378060479.55 | TIME:   0.004667
Time of first run  0.010165161999793781
ERF: Numba@jit-loop-par | Size:   67108864 | MOPS:  16862131587.41 | TIME:   0.007960
Time of first run  0.017497161999926902
ERF: Numba@jit-loop-par | Size:  134217728 | MOPS:  19211915654.84 | TIME:   0.013972
Time of first run  0.03390368500004115
ERF: Numba@jit-loop-par | Size:  268435456 | MOPS:  20912204492.02 | TIME:   0.025673

numba_dpex/core/kernel_interface/dispatcher.py
```diff
@@ -626,12 +632,25 @@ def __call__(self, *args):
...
-        # create a sycl::KernelBundle
-        kernel_bundle = dpctl_prog.create_program_from_spirv(
-            exec_queue,
-            device_driver_ir_module,
-            " ".join(self._create_sycl_kernel_bundle_flags),
+        kernel_bundle_key = build_key(
+            tuple(argtypes),
+            self.pyfunc,
+            dpex_kernel_target.target_context.codegen(),
+            backend=backend,
```
diptorupd, line 639: The key for the `kernel_bundle` has to store the queue object and not `backend` and `device_type`. A kernel bundle is tied to a SYCL context and having the same device_type and backend does not guarantee that the context is the same. If we pass a kernel_bundle that was compiled for a specific context to a queue on a different context that will lead to a crash.

https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_the_kernel_bundle_class

diptorupd: Regarding the point raised by [fcharras](https://github.com/fcharras), do we really need to hash the codegen as part of the cache key? It may be adding some extra overhead.

numba_dpex/core/kernel_interface/dispatcher.py
chudur-budur: Yea it's only for debugging, but we use it in unit tests though.

adarshyoga: Can we then just use it in debug mode by wrapping it in "if config.DEBUG_CACHE:"? Although a single update overhead can add up.

numba_dpex/core/kernel_interface/dispatcher.py
```diff
@@ -95,6 +95,11 @@ def __init__(
                 capacity=config.CACHE_SIZE,
                 pyfunc=self.pyfunc,
             )
+            self._kernel_bundle_cache = LRUCache(
+                name="KernelBundleCache",
+                capacity=config.CACHE_SIZE,
+                pyfunc=self.pyfunc,
+            )
```
adarshyoga, line 102: The _kernel_bundle_cache needs to be defined in the if and else blocks to and set to NullCache.

adarshyoga (changes requested):

.pre-commit-config.yaml
diptorupd: I am pushing this isort change to main. Without the version upgrade, pre-commit stopped working.

numba_dpex/core/kernel_interface/dispatcher.py
diptorupd: I doubt updating the flag is having any discernible overhead.

numba_dpex/core/kernel_interface/dispatcher.py
```diff
@@ -625,12 +632,28 @@ def __call__(self, *args):
...
-        # create a sycl::KernelBundle
-        kernel_bundle = dpctl_prog.create_program_from_spirv(
-            exec_queue,
-            device_driver_ir_module,
-            " ".join(self._create_sycl_kernel_bundle_flags),
+        kernel_bundle_key = build_key(
+            tuple(argtypes),
+            self.pyfunc,
+            dpex_kernel_target.target_context.codegen(),
+            # backend=backend,
```
diptorupd, line 639: Please remove commented lines.

numba_dpex/tests/kernel_tests/test_arg_types.py
```diff
@@ -53,6 +55,9 @@ def check_bool_kernel(A, test):
         A[0] = 222
 
 
+@pytest.mark.skipif(
```
diptorupd, line 58: We have to investigate it further, in the past we have had code generation issues caused due to improper pointer width for 64bit ints getting generated on Window. 

As a first step, can you convert the test to use dpnp/dpctl.tensor and compute follows data. If the issue is not showing up with dpnp/dpctl.tensor we are golden.

numba_dpex/core/caching.py
```diff
@@ -13,7 +13,9 @@
 from numba_dpex.core.types import USMNdArray
 
 
-def build_key(argtypes, pyfunc, codegen, backend=None, device_type=None):
+def build_key(
+    argtypes, pyfunc, codegen, backend=None, device_type=None, exec_queue=None
+):
     """Constructs a key from python function, context, backend and the device
```
diptorupd, line 19: Fix docstring to indicate changes.

numba_dpex/tests/kernel_tests/test_arg_types.py
```diff
@@ -37,13 +40,39 @@ def input_arrays(request):
...
+        b.shape,
+        dtype=b.dtype,
+        buffer=usm_type,
+        buffer_ctor_kwargs={"queue": queue},
+    )
+    db.usm_data.copy_from_host(b.reshape((-1)).view("|u1"))
+
+    kernel = dpex.kernel(mul_kernel)
+    kernel[dpex.NdRange(dpex.Range(global_size), dpex.Range(local_size))](
+        da, db, c
```
diptorupd, line 69: I am surprised that this code works. We are mixing tensor and numpy. @chudur-budur  please check why an exception is not raised here.

`c` should be a numpy array, in fact you can just pass a scalar value. The kernel is only scaling  `a` into `b`.

numba_dpex/tests/kernel_tests/test_arg_types.py
```diff
@@ -37,13 +40,39 @@ def input_arrays(request):
...
 @pytest.mark.parametrize("filter_str", filter_strings)
 def test_kernel_arg_types(filter_str, input_arrays):
-    kernel = dpex.kernel(mul_kernel)
-    a, actual, c = input_arrays
+    usm_type = "device"
+
+    a, b, c = input_arrays
     expected = a * c
+
     device = dpctl.SyclDevice(filter_str)
```
diptorupd, line 48: We want to get away from this pattern as it leads to lots of skipped tests when the filter string is not supported. Instead of using the filter string parameter, just set the device to `dpctl.select_default_device()`. In future, we will remove all filter string params and control default device using SYCL environment variables.

numba_dpex/tests/kernel_tests/test_arg_types.py
```diff
@@ -55,12 +84,28 @@ def check_bool_kernel(A, test):
...
+        a.shape,
+        dtype=a.dtype,
+        buffer=usm_type,
+        buffer_ctor_kwargs={"queue": queue},
+    )
+    da.usm_data.copy_from_host(a.reshape((-1)).view("|u1"))
+
+    kernel = dpex.kernel(check_bool_kernel)
+
+    kernel[dpex.Range(a.size)](da, True)
```
diptorupd, line 103: This test can also be done using a scalar and setting range to `1`, in effect launching a sycl single_task.

diptorupd: Also, you can directly use dpctl constructors and do not need to copy data from NumPy.

diptorupd:
@chudur-budur Looks good! Need some minor changes to the test case. Can you also squash your changes into just two commits: one with the caching changes and the other with the test case changes.

chudur-budur:
> @chudur-budur Looks good! Need some minor changes to the test case. Can you also squash your changes into just two commits: one with the caching changes and the other with the test case changes.

Thanks! I think it's good to go.

diptorupd:
@ fcharras Can you please evaluate the PR to see if you see performance improvements in your benchmarks?

fcharras:
After https://github.com/IntelPython/numba-dpex/issues/898 my benchmarks work again and I can compare properly performances before and after #804 .

This PR definitely fixes a major slowdown that would have made the JIT near unusable for me, it's a :+1: for merge

After this PR the performances before and after #804 have the same order of magnitude and is usable again.

But the python overhead is not as negligible as it used to be before #804 . I would not say it completely fixes https://github.com/IntelPython/numba-dpex/issues/886 . As suggested there, maybe the caching instructions can be made a bit more efficient (e.g attaching the codegen to the jitkernel instances). (or any other bottleneck that a profiler can highlight).

The added overhead I see in our KMeans is about 1sc per run, for small workload it's noticeable.

diptorupd (approved):

PR #899, opened 2023-02-03
Make alias of Range/NdRange in __init__

chudur-budur:
Now instead of doing `from numba_dpex.core.kernel_interface.utils import Range, NdRange` you can import them like `from numba_dpex import Range, NdRange`

### DIFF
--- a/numba_dpex/examples/kernel/scan.py
+++ b/numba_dpex/examples/kernel/scan.py
@@ -7,7 +7,6 @@
 import dpnp as np
 
 import numba_dpex as ndpx
-from numba_dpex.core.kernel_interface.utils import Range
 
 # 1D array size
 N = 64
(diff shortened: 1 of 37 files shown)


numba_dpex/examples/kernel/scan.py
```diff
@@ -7,7 +7,6 @@
 import dpnp as np
 
 import numba_dpex as ndpx
```
adarshyoga, line 9: Better to keep the import statements consistent in all examples. If were doing "import numba_dpex as dpex" in other examples, we should do the same here.

diptorupd (approved):
LGTM! Please rebase on top of main and merge.

numba_dpex/examples/kernel/scan.py
chudur-budur: Done.

chudur-budur:
> LGTM! Please rebase on top of main and merge.

Done. Thanks!

PR #900, opened 2023-02-03
Updates the black requirement to black 23.1

diptorupd:
Numba-dpex now uses black 23.1 and the PR updates the pre-commit config and reformat the files impacted by the change.


diptorupd:
Black and flake8 are in contradiction. We have to pick one!

PR #911, opened 2023-02-15
First attempt

AlexanderKalistratov:
* Doesn't work.
* Created native function (`DPEXRT_get_queue_ref`) which I think extracts native pointer to sycl queue. It is exposed to Python as integer and could be captured by Numba jitted function during typing stage.
* Add `_data_ptr` accessor to DpnpNdArray to access underlying data from jitted functions
* Updated enum with DPNP functions
* Change mapping of string `"dpnp_cumsum"` from `DPNP_FN_CUMSUM` to `DPNP_FN_CUMSUM_EXT`
* Current problem is what signature of dpnp function `(int, void*, void*, int, int)->void*` doesn't match actually passed types `(int, float*, dtype(float)*, int, int)`. Not sure if `Numba` a has `float*` type. Either need to try to bind with `ctypes` signature or use `intrinsic` to do cast/call.


AlexanderKalistratov:
@diptorupd @mingjie-intel

numba_dpex/dpnp_iface/dpnp_array_ops_impl.py
```diff
@@ -47,26 +46,49 @@ def common_impl(a, out, dpnp_func, print_debug):
...
+    res_type = types.voidptr
     """
     dpnp source:
     https://github.com/IntelPython/dpnp/blob/0.5.1/dpnp/backend/kernels/dpnp_krnl_mathematical.cpp#L135
     Function declaration:
     void dpnp_cumsum_c(void* array1_in, void* result1, size_t size)
     """
-    sig = signature(res_type, types.voidptr, types.voidptr, types.intp)
+    # sig = signature(res_type, types.voidptr, types.voidptr, types.intp)
+    sig = signature(res_type, types.intp, types.voidptr, types.voidptr, types.intp, types.intp)
```
AlexanderKalistratov, line 79: This signature doesn't match actualy passed types (void* vs float*)

numba_dpex/core/datamodel/models.py
```diff
@@ -86,3 +86,4 @@ def _init_data_model_manager():
 # Register the DpctlSyclQueue type with Numba's OpaqueModel
 register_model(DpctlSyclQueue)(OpaqueModel)
 dpex_data_model_manager.register(DpctlSyclQueue, OpaqueModel)
+make_attribute_wrapper(DpnpNdArray, 'data', '_data_ptr')
```
AlexanderKalistratov, line 89: Makes `data` from model accessible from jitted functions

numba_dpex/dpnp_iface/dpnp_array_ops_impl.py
```diff
@@ -47,26 +46,49 @@ def common_impl(a, out, dpnp_func, print_debug):
...
     dpnp_func = dpnp_ext.dpnp_func("dpnp_" + name, [a.dtype.name, "NONE"], sig)
+    sycl_queue = numba_dpex.core.runtime.get_queue_ref(a.queue)
```
AlexanderKalistratov, line 81: Extracts native pointer to sycl queue as python integer. Probably.

PR #912, opened 2023-02-15
Dispatcher/caching rewrite to address performance regression

adarshyoga:
This PR partially addresses the performance regressions described in #886. It contains the following 4 key changes.

(1) ~~The put() function implementation in LRUCache contains a call to get(), which was unnecessary. This PR removes the call to get().~~ Replaced call to get() in LRUCache.put() method with explicit logic to update the linked list keeping track of LRU ordering.

(2) Sha256 hash computation was being performed on every call to build cache key, which in turn was being called on for each dynamic call to a kernel. Instead of computing hash on every dynamic call to a kernel, it can be done once. This PR performs the hash computation for every static instance of a kernel rather than every dynamic instance.

(3) The types of kernel arguments are used as a part of the key when caching. The arguments types were being pre-processed to strip out USM metadata. This functionality was being performed twice for each call, once for building key for kernel module cache and again for kernel bundle cache. This PR changes the logic to perform the pre-processing once per-call.

(4) The function that build the cache key takes variable number of arguments and returns a tuple. The rest of the logic from build key has been moved to dispatcher and func. The key intuition of using variable args is to support the different caches, so far, kernel module cache and kernel bundle cache. Both these caches use different number of keys. (Side note: The build_key function is ideally suited to exist as a static method inside AbstractCache class).

**Effects of optimizations:**
Evaluated the effect of these changes using kmeans implementation from [here](https://github.com/soda-inria/sklearn-numba-dpex).
On Intel ATS GPUs the execution time before the changes is 1.7 seconds. After these changes the execution time reduces to 1.4 seconds. See log below. With numba-dpex 0.19.0 the execution time is 1.1 seconds. These changes partially address the regression introduced after 0.19.0.

Run Log with numba-dpex 0.19.0:
-------------------------------------
python benchmark/kmeans.py
Running Kmeans numba_dpex lloyd GPU ... done in 1.1 s

Run Log with numba-dpex main:
-------------------------------------
python benchmark/kmeans.py
Running Kmeans numba_dpex lloyd GPU ... done in 1.7 s

Run Log with this PR:
------------------------
python benchmark/kmeans.py
Running Kmeans numba_dpex lloyd GPU ... done in 1.4 s

### DIFF
--- a/numba_dpex/core/caching.py
+++ b/numba_dpex/core/caching.py
@@ -2,77 +2,16 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import hashlib
 import sys
 from abc import ABCMeta, abstractmethod
 
 from numba.core.caching import CacheImpl, IndexDataCacheFile
-from numba.core.serialize import dumps
 
 from numba_dpex import config
-from numba_dpex.core.types import USMNdArray
 
 
-def build_key(
-    argtypes, pyfunc, codegen, backend=None, device_type=None, exec_queue=None
-):
-    """Constructs a key from python function, context, backend, the device
-    type and execution queue.
-
-    Compute index key for the given argument types and codegen. It includes a
-    description of the OS, target architecture and hashes of the bytecode for
-    the function and, if the function has a __closure__, a hash of the
-    cell_contents.type
-
-    Args:
-        argtypes : A tuple of numba types corresponding to the arguments to the
-        compiled function.
-        pyfunc : The Python function that is to be compiled and cached.
-        codegen (numba.core.codegen.Codegen):
-            The codegen object found from the target context.
-        backend (enum, optional): A 'backend_type' enum.
-            Defaults to None.
-        device_type (enum, optional): A 'device_type' enum.
-            Defaults to None.
-        exec_queue (dpctl._sycl_queue.SyclQueue', optional): A SYCL queue object.
-
-    Returns:
-        tuple: A tuple of return type, argtpes, magic_tuple of codegen
-            and another tuple of hashcodes from bytecode and cell_contents.
-    """
-
-    codebytes = pyfunc.__code__.co_code
-    if pyfunc.__closure__ is not None:
-        try:
-            cvars = tuple([x.cell_contents for x in pyfunc.__closure__])
-            # Note: cloudpickle serializes a function differently depending
-            #       on how the process is launched; e.g. multiprocessing.Process
-            cvarbytes = dumps(cvars)
-        except:
-            cvarbytes = b""  # a temporary solution for function template
-    else:
-        cvarbytes = b""
-
-    argtylist = list(argtypes)
-    for i, argty in enumerate(argtylist):
-        if isinstance(argty, USMNdArray):
-            # Convert the USMNdArray to an abridged type that disregards the
-            # usm_type, device, queue, address space attributes.
-            argtylist[i] = (argty.ndim, argty.dtype, argty.layout)
-
-    argtypes = tuple(argtylist)
-
-    return (
-        argtypes,
-        codegen.magic_tuple(),
-        backend,
-        device_type,
-        exec_queue,
-        (
-            hashlib.sha256(codebytes).hexdigest(),
-            hashlib.sha256(cvarbytes).hexdigest(),
-        ),
-    )
+def build_key(*args):
+    return tuple(args)
 
 
 class _CacheImpl(CacheImpl):
@@ -476,7 +415,6 @@ class LRUCache(AbstractCache):
                     )
                 )
             self._lookup[key].value = value
-            self.get(key)
             return
 
         if key in self._evicted:
--- a/numba_dpex/core/caching.py
+++ b/numba_dpex/core/caching.py
@@ -10,10 +10,6 @@ from numba.core.caching import CacheImpl, IndexDataCacheFile
 from numba_dpex import config
 
 
-def build_key(*args):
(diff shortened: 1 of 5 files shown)


numba_dpex/core/caching.py
```diff
@@ -2,77 +2,16 @@
...
-        codegen.magic_tuple(),
-        backend,
-        device_type,
-        exec_queue,
-        (
-            hashlib.sha256(codebytes).hexdigest(),
-            hashlib.sha256(cvarbytes).hexdigest(),
-        ),
-    )
+def build_key(*args):
```
diptorupd, line 13: Please add a docstring.

numba_dpex/core/kernel_interface/dispatcher.py
```diff
@@ -143,6 +147,28 @@ def __init__(
             self._has_specializations = False
             self._specialization_cache = NullCache()
 
+    def _create_func_hash(self):
+        """Creates a tuple of sha256 hashes out of code and variable bytes extracted from the compiled funtion."""
```
diptorupd, line 151: Docstring is more the 70 chars. Please add line breaks.

numba_dpex/core/caching.py
diptorupd: Why even have this call? We can as well create the tuple where needed and insert or lookup into the cache.

numba_dpex/core/kernel_interface/dispatcher.py
```diff
@@ -586,6 +614,20 @@ def _check_ranges(self, device):
                 device=device,
             )
 
+    def _strip_usm_metadata(self, argtypes):
+        stripped_argtypes = []
+        for argty in argtypes:
+            if isinstance(argty, USMNdArray):
+                # Convert the USMNdArray to an abridged type that disregards the
```
diptorupd, line 621: Might as well move this comment into function docstring.

numba_dpex/core/kernel_interface/func.py
```diff
@@ -103,6 +108,28 @@ def __init__(self, pyfunc, debug=False, enable_cache=True):
             self._cache = NullCache()
         self._cache_hits = 0
 
+    def _create_func_hash(self):
+        """Creates a tuple of sha256 hashes out of code and variable bytes extracted from the compiled funtion."""
```
diptorupd, line 112: Instead of duplicating the function in both dispatcher and func, how about keeping it in caching.py and calling it in both places?

numba_dpex/core/kernel_interface/func.py
```diff
@@ -113,6 +140,20 @@ def cache_hits(self):
         """Cache hit count accessor"""
         return self._cache_hits
 
+    def _strip_usm_metadata(self, argtypes):
```
diptorupd, line 143: Same with this function, can be moved to a common place and reused.

diptorupd:
Thanks @adarshyoga for triaging it. 

The `func_hash` is really needed if we run into a situation where the number of cached versions of a kernel exceeds `NUMBA_DPEX_CACHE_SIZE`. Note that the cache limit is per kernel. If the limit is exceeded, then we will start to pickle the kernel to disk. 

Can you try making the func_hash optional based on a flag? That can save us some cycles.

numba_dpex/core/kernel_interface/dispatcher.py
```diff
@@ -171,11 +197,13 @@ def _compile_and_cache(self, argtypes, cache):
...
         kernel_module_name = kernel.module_name
 
-        key = build_key(
-            tuple(argtypes),
-            self.pyfunc,
-            kernel.target_context.codegen(),
-        )
+        if not key:
+            stripped_argtypes = self._strip_usm_metadata(argtypes)
+            codegen_magic_tuple = kernel.target_context.codegen().magic_tuple()
```
diptorupd, line 202: I compared with 0.19.0, even this extra hashing on the codegen_magic_tuple may be adding an overhead. Can you try removing this and just adding the kernel to the key like we had in 0.19

chudur-budur (changes requested):

numba_dpex/core/kernel_interface/dispatcher.py
adarshyoga: Done

numba_dpex/core/kernel_interface/dispatcher.py
adarshyoga: Done

numba_dpex/core/kernel_interface/func.py
adarshyoga: Moved to separate caching_utils.py

numba_dpex/core/kernel_interface/func.py
adarshyoga: Moved to separate caching_utils.py

numba_dpex/core/caching.py
```diff
@@ -476,7 +415,6 @@ def put(self, key, value):
                     )
                 )
             self._lookup[key].value = value
-            self.get(key)
```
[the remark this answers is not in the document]
adarshyoga: I see. I have replaced get() with explicit logic to update list that maintain LRU ordering. Ran the test case after the change. It produces same output as when get() was being used. 
Hope this addresses you comment.

chudur-budur (approved):

chudur-budur:
LGTM! Looks like we need to manually restart those stuck jobs at teamcity.

PR #920, opened 2023-02-17
Revise dpnp overloads

AlexanderKalistratov:
* Implement `get_queue_ref` to extract native sycl queue pointer to pass it to dpnp functions
* Update `cumsum` overload as an example
* Implemented own native version of `cumsum` to test if all parameters are passed correctly from numba to native. (for debbuging purpose; to be removed)
* Added example to quickly test cumsum (for debbuging purpose; to be removed)

```
cd example
icpx -fPIC -fsycl -shared -o testlib.so testlib.cpp
python ./example.py
```

Issues:
* `x.cumsum()` doesn't work. `dpnp.cumsum(x)` does.
* `dpnp.cumsum(x)` doesn't work. `dpnp.cumsum(x, out)` does.
* Waiting for event inside jitted function doesn't work. Waiting for event inside native function works.


AlexanderKalistratov:
@diptorupd @mingjie-intel

Issue #906, opened 2023-02-11
Minimal reproducer for incorrect code generated for multiple barriers followed by multiple conditional ops in a kernel


fcharras:
The following snippet:

```
import numba_dpex as dpex
import dpctl.tensor as dpt
import numpy as np

dtype = np.float32


@dpex.kernel
def kernel(result):
    local_col_idx = dpex.get_local_id(0)
    local_values = dpex.local.array((1,), dtype=dtype)

    dpex.barrier(dpex.LOCAL_MEM_FENCE)

    if (local_col_idx < 1):
        local_values[0] = 1

    dpex.barrier(dpex.LOCAL_MEM_FENCE)

    if (local_col_idx < 1):
        result[0] = 10

result = dpt.zeros(sh=(1), dtype=dtype)
kernel[32, 32](result)
print(result)

```

when **ran on CPU**, prints:
```
[0.]
```

but it should print:
```
[10.]
```

I think it is a simpler instance of https://github.com/IntelPython/numba-dpex/issues/892 . The buggy pattern seems to be this particular sequence of instruction:  barrier -> conditional op on local memory -> barrier -> conditional op on global memory. You can see how the correct result is printed when one of this step is altered.

Also, it does not occur for all group sizes. Here, `[32, 32]` triggers the bug but `[16, 16]` works. (if you can't reproduce the issue, maybe try higher group sizes)

I can't reproduce it on GPU but maybe there's a combination of group size that could make it fail too.

mingjie-intel:
An update:

1. With reproducer, 0 can be reproduced on CPU with dpex.  
2. A dpcpp code is written to reproduce the issue. 10 is reported both on cpu and gpu. 

```
#include <CL/sycl.hpp>

using namespace sycl;

static const int N = 1;

int main() {

  std::array<float, N> host_a;

  for(int i=0; i<N; i++)
    host_a[i]=0;

  queue q(cpu_selector_v);
  //queue q;
  std::cout << "Device : " << q.get_device().get_info<info::device::name>() << "\n";

  float *device_a = malloc_device<float>(N, q);

  range global{32};
  range local{32};
  q.parallel_for(nd_range{global, local}, [=](nd_item<1> idx) {
                                size_t i = idx.get_global_id(0);
                                float local_val = 0.;
                                idx.barrier(sycl::access::fence_space::local_space);
                                if(i<1) local_val=1.;
                                idx.barrier(sycl::access::fence_space::local_space);
[...]

PR #924, opened 2023-02-21
Output LLVM IR if NUMBA_DPEX_DEBUG environment variable is set

adarshyoga:

Adding functionality to output LLVM IR to a file. The output file name contains a hash generated from the LLVM module name. This allows us to write different kernels to separate files since the hash is expected to be unique.

Tested with Blackscholes in dpbench. File name generated is `llvm_kernel_ce84b7ebc25a6c5f71403d77de797be75067dce745a14ae28ca34dc5e3365998.ll`

Tested on GPU

### DIFF
--- a/numba_dpex/core/kernel_interface/spirv_kernel.py
+++ b/numba_dpex/core/kernel_interface/spirv_kernel.py
@@ -7,7 +7,7 @@ from types import FunctionType
 
 from numba.core import ir
 
-from numba_dpex import spirv_generator
+from numba_dpex import config, spirv_generator
 from numba_dpex.core.compiler import compile_with_dpex
 from numba_dpex.core.exceptions import UncompiledKernelError, UnreachableError
 
@@ -138,6 +138,20 @@ class SpirvKernel(KernelInterface):
         )
         self._llvm_module = kernel.module.__str__()
         self._module_name = kernel.name
+
+        # Dump LLVM IR if DEBUG flag is set.
+        if config.DEBUG:
+            import hashlib
+
+            # Embed hash of module name in the output file name
+            # so that different kernels are written to separate files
+            with open(
+                "llvm_kernel_"
+                + hashlib.sha256(self._module_name.encode()).hexdigest()
+                + ".ll",
+                "w",
+            ) as f:
+                f.write(self._llvm_module)
 
         # FIXME: There is no need to serialize the bitcode. It can be passed to
         # llvm-spirv directly via stdin.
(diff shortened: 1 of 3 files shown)


numba_dpex/core/kernel_interface/spirv_kernel.py
```diff
@@ -139,6 +139,20 @@ def compile(
...
         self._module_name = kernel.name
 
+        # Dump LLVM IR if DEBUG flag is set.
+        if config.DEBUG:
+            import hashlib
+
+            # Embed hash of module name in the output file name
+            # so that different kernels are written to separate files
+            with open(
+                "llvm_kernel_"
```
diptorupd, line 149: We can use `self._module_name` in the file name so that it is clear. Although, with the type of kernel names we generate for parfors that might be exposing skeletons on the closet 😄

numba_dpex/core/kernel_interface/spirv_kernel.py
```diff
@@ -139,6 +139,20 @@ def compile(
         self._llvm_module = kernel.module.__str__()
         self._module_name = kernel.name
 
+        # Dump LLVM IR if DEBUG flag is set.
+        if config.DEBUG:
```
diptorupd, line 143: Can we use a new flag `NUMBA_DPEX_DUMP_KERNEL_LLVM`? The `DEBUG` seems to turn on debug symbols in some places.

numba_dpex/core/kernel_interface/spirv_kernel.py
diptorupd: :speak_no_evil:

mingjie-intel:
Tested the code with small code from #906 , llvm IR is printed.

numba_dpex/core/kernel_interface/spirv_kernel.py
adarshyoga: Added. Could not find any place where documentation needs to be updated with the env var. If I have missed it, let me know.

diptorupd:
@adarshyoga Thank you! It is good to go, but we need a test case. Just to verify that setting the config actually generates the IR file in the right location and the file is not empty.

Refer: https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmpdir-fixture

adarshyoga:
> @adarshyoga Thank you! It is good to go, but we need a test case. Just to verify that setting the config actually generates the IR file in the right location and the file is not empty.
> 
> Refer: https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmpdir-fixture

Added a test case that tests both positive and negative cases.

PR #926, opened 2023-02-21
Fixed a blunder in _parse_version

oleksandr-pavlyk:
Added hastily remove slicing of dpctl version tuple to the first two elements which are expected to be integral.

### DIFF
--- a/numba_dpex/dpctl_support.py
+++ b/numba_dpex/dpctl_support.py
@@ -9,7 +9,7 @@ def _parse_version():
     t = dpctl.__version__.split(".")
     if len(t) > 1:
         try:
-            return tuple(map(int, t))
+            return tuple(map(int, t[:2]))
         except ValueError:
             return (0, 0)
     else:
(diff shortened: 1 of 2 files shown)


oleksandr-pavlyk:
Perhaps a test is in order to verify that `numba_dpex.dpctl_support.dpctl_version` is a 2-tuple with integral values consistent with the `dpctl.__version__`. Something like

```python
def test_dpctl_version():
    dpctl_v = dpctl.__version__
    computed_v = ".".join(dpctl_version)
    assert len(computed_v) <= len(dpctl_v)
    assert computed_v == dpctl_v[:len(computed_v)]
```

Where would such a test belong?

diptorupd:
> Perhaps a test is in order to verify that `numba_dpex.dpctl_support.dpctl_version` is a 2-tuple with integral values consistent with the `dpctl.__version__`. Something like
> 
> ```python
> def test_dpctl_version():
>     dpctl_v = dpctl.__version__
>     computed_v = ".".join(dpctl_version)
>     assert len(computed_v) <= len(dpctl_v)
>     assert computed_v == dpctl_v[:len(computed_v)]
> ```
> 
> Where would such a test belong?

Please add to `numba_dpex/tests` in the top level folder.

diptorupd:
thank you!

diptorupd (approved):

PR #928, opened 2023-02-24
Implementation for dpnp.empty_like(), dpnp.zeros_like() and dpnp.ones_like()

chudur-budur:
1. Implementation for overloading `dpnp.empty_like()`, `dpnp.zeros_like()` and `dpnp.ones_like()` type of functions.

2. This also does code-reorganization in [`arrayobj.py`](https://github.com/IntelPython/numba-dpex/blob/main/numba_dpex/dpnp_iface/arrayobj.py)


chudur-budur:
Yea removed `shape` supports, it's good to go now. @diptorupd

diptorupd:
@chudur-budur Thanks!

It is interesting that we can create a dpnp array from these new constructors from a NumPy array. Does dpnp even support such a feature.

I am merging as is, we can revisit later if needed.

PR #938, opened 2023-02-28
Feature/dpnp parfor

diptorupd:

Adds support for parfor lowering for dpnp.ndarrays.

- Updates to DpnpNdArray type to support compute follows data inside dpjit decorated functions
- Back ported the new parfor passes from Numba
- New parfor lowerer that supports dpnp ndarrays and removes all extra memcpy functions at parfor boundaries
- Fixes and clean up of unused features from the existing lowerer._gufunc_parfor_lowerer
- New dpjit compiler pipeline that uses the new parfor passes.
- Switched the dpjit decorator to use the dpjit pipeline instead of the offload pipeline

- [X] If this PR is a work in progress, are you filing the PR as a draft?


diptorupd:
@oleksandr-pavlyk what do you say about this coverage drop!!

oleksandr-pavlyk:
@diptorupd Ok, coverage drop of 2% in `dpctl` pales in comparison. A lot more tests are needed

PR #939, opened 2023-02-28
Use dpnp ext iface for cumsum

AlexanderKalistratov:
Example of using dpnp ext interface for cumsum function


diptorupd:
Closing the PR as the `dpnp_iface` module is getting redesigned and the dpnp backend module is deprecated.

Issue #759, opened 2022-07-28
wrong output dtype for math functions


fcharras:
`math.ceil` and `math.floor` [are supported within a numba kernel](https://intelpython.github.io/numba-dpex/latest/user_guides/kernel_programming_guide/supported-python-features.html#standard-library-modules) and when using it within a `CPython` interpreter it returns an `int`.

But within a `dpex.kernel` it returns a `float64`, which can be the source of several issues:

- the output of those `math` functions cannot be used to index a `dpex.local.array`
- it will fail to compile when the gpu do not have `float64` support

Here is a toy example to reproduce the issue:
```
 import dpctl
import numpy as np
from numba import float32, uint32
import numba_dppy as dpex
import math

def get_kernel(p, q, r):
    shm_shape = (p + 1, q + 1)
    @dpex.kernel
    def test_kernel(a, b):
        threadx = dpex.get_global_id(0)
        thready = dpex.get_global_id(1)
        some_shared_mem = dpex.local.array(shape=shm_shape, dtype=float32)
        x = math.ceil(1 + threadx // r)
        y = math.ceil(1 + thready // r)
        some_shared_mem[x, y] = float32(3.)
        b[threadx, thready] = a[threadx, thready] + float32(3.) + float32(math.inf)
    return test_kernel

# Array dimensions
X = 16 * 16
Y = 16
global_size = X, X

griddim = X, X
blockdim = Y, Y

a = np.arange(X * X, dtype=np.float32).reshape(X, X)
b = np.empty_like(a)

# Use the environment variable SYCL_DEVICE_FILTER to change the default device.
# See https://github.com/intel/llvm/blob/sycl/sycl/doc/EnvironmentVariables.md#sycl_device_filter.
device = dpctl.select_default_device()
print("Using device ...")
device.print_device_info()

with dpctl.device_context(device):
    get_kernel(4, 4, 4)[griddim, blockdim](a, b)

print("Done...")
```

It fails both on Intel devcloud or a [custom environment](https://github.com/ogrisel/sklearn-numba-dpex/blob/main/docker/Dockerfile) with more up to date versions

chudur-budur:
Minimal reproducer using latest syntax:
```python
def test_minimal():
    N = 10
    device = dpctl.select_default_device()
    
    @dpx.kernel
    def func(a):
        i = dpx.get_global_id(0)
        i = math.ceil(i + 0.5) - 1
        a[i] = a[i] + 1
    
[...]

PR #960, opened 2023-03-06
Fix issue with math.ceil() and math.floor() output

chudur-budur:
Addressing issue #759 

`math.ceil()`  and `math.floor()` are supposed to return `int` inside `numba_dpex.kernel`, but they are returning `float64`. This PR should fix that issue.

### DIFF
--- a/numba_dpex/ocl/mathdecl.py
+++ b/numba_dpex/ocl/mathdecl.py
@@ -230,10 +230,24 @@ class Math_atanh(Math_unary):
 
 class Math_floor(Math_unary):
     key = math.floor
+    cases = [
+        signature(types.intp, types.intp),
+        signature(types.int64, types.int64),
+        signature(types.uint64, types.uint64),
+        signature(types.int64, types.float32),
+        signature(types.int64, types.float64),
+    ]
 
 
 class Math_ceil(Math_unary):
     key = math.ceil
+    cases = [
+        signature(types.intp, types.intp),
+        signature(types.int64, types.int64),
+        signature(types.uint64, types.uint64),
+        signature(types.int64, types.float32),
+        signature(types.int64, types.float64),
+    ]
 
 
 class Math_trunc(Math_unary):
--- a/numba_dpex/tests/kernel_tests/test_math_functions.py
+++ b/numba_dpex/tests/kernel_tests/test_math_functions.py
@@ -5,13 +5,24 @@
 import math
 
 import dpctl
+import dpnp
 import numpy as np
 import pytest
 
 import numba_dpex as dpex
 from numba_dpex.tests._helper import filter_strings
 
-list_of_unary_ops = ["fabs", "exp", "log", "sqrt", "sin", "cos", "tan"]
+list_of_unary_ops = [
+    "fabs",
+    "exp",
+    "log",
+    "sqrt",
+    "sin",
+    "cos",
+    "tan",
+    "ceil",
+    "floor",
+]
 
 
 @pytest.fixture(params=list_of_unary_ops)
@@ -24,31 +35,44 @@ list_of_dtypes = [
     np.float64,
 ]
 
+N = 2048
+
 
 @pytest.fixture(params=list_of_dtypes)
 def input_arrays(request):
     # The size of input and out arrays to be used
-    N = 2048
     a = np.array(np.random.random(N), request.param)
     b = np.array(np.random.random(N), request.param)
-    return a, b
+    c = np.zeros(N, dtype=np.int64)
+    return a, b, c
 
 
 @pytest.mark.parametrize("filter_str", filter_strings)
 def test_binary_ops(filter_str, unary_op, input_arrays):
-    a, actual = input_arrays
+    a, actual, actual_value_types = input_arrays
+
+    if unary_op == "ceil" or unary_op == "floor":
+        a = a * 10.0
+
     uop = getattr(math, unary_op)
     np_uop = getattr(np, unary_op)
 
     @dpex.kernel
-    def f(a, b):
+    def f(a, b, c):
         i = dpex.get_global_id(0)
-        b[i] = uop(a[i])
+        k = uop(a[i])
+        if type(k) is int or type(k) is np.int32 or type(k) is np.int64:
+            c[i] = 3
+        b[i] = k
 
     device = dpctl.SyclDevice(filter_str)
     with dpctl.device_context(device):
-        f[a.size, dpex.DEFAULT_LOCAL_SIZE](a, actual)
+        f[a.size, dpex.DEFAULT_LOCAL_SIZE](a, actual, actual_value_types)
 
     expected = np_uop(a)
 
     np.testing.assert_allclose(actual, expected, rtol=1e-5, atol=0)
+
+    if unary_op == "ceil" or unary_op == "floor":
+        expected_value_types = (np.ones(N) * 3).astype(np.int64)
+        np.testing.assert_equal(actual_value_types, expected_value_types)


AlexanderKalistratov:
Please note this is a difference between `c++/sycl/ocl` and `python`.
In `c++/sycl/ocl` `ceil` always returns floating pointer type. In `python` it returns `int`

chudur-budur:
> Please note this is a difference between `c++/sycl/ocl` and `python`. In `c++/sycl/ocl` `ceil` always returns floating pointer type. In `python` it returns `int`

Exactly, `ceil()`,`floor()` in all `c/c++` implementations out there do `float --> float` or `double --> double`, including SYCL. We are discussing with inria what we should do, whether we follow the `python` construct or `c/c++` , we are discussing here in issue #759

fcharras:
@chudur-budur I can confirm that on this branch `math.ceil` returns an int.

But, looking back at https://github.com/IntelPython/numba-dpex/issues/759 , the most troublesome point was that `math.ceil` and `math.floor` do not to work with devices that do not support `float64` compute (such as igpus), and this PR does not solve that point. 

Even if casting everything `float32` beforehand, e.g:

```python
import dpctl.tensor as dpt
import dpnp
import numba_dpex as dpx
import math

N = 10

cst = dpnp.float32(0.5)

@dpx.kernel
def func(a):
    i = dpx.get_global_id(0)
    i = math.ceil(dpnp.float32(i + cst))
    a[i] = a[i] + i

a = dpt.ones(N, dtype=dpnp.float32)

func[dpx.Range(N)](a)

print(dpt.asnumpy(a))
```

gives a `SyclProgramCompilationError:` which is the error we're used to for `float64` kernels on `float32`-only  hardware.

fcharras:
In fact forget what I just said, this works:

```python
import dpctl.tensor as dpt
import dpnp
import numba_dpex as dpx
import math

N = 10

cst = dpnp.float32(0.5)

@dpx.kernel
def func(a):
    i = dpx.get_global_id(0)
    i = math.ceil(dpnp.float32(i) + cst)
    a[i] = a[i] + dpnp.float32(i)

a = dpt.ones(N, dtype=dpnp.float32)

func[dpx.Range(N)](a)

print(dpt.asnumpy(a))
```

so the remaining issue in this example was that `a + b` fails to compile if `a` is `int64` but `b` is `float32`, explicitly casting `a` to `float32` before summing fixes the issue. Without the explicit casting, it seems the JIT casts `a` from `int64` to `float64`.

Knowing this is probably enough to write `numba_dpex` kernels that can work on all devices but one must think of casting explicitly to have compatibility with igpus.

chudur-budur:
> so the remaining issue in this example was that `a + b` fails to compile if `a` is `int64` but `b` is `float32`, explicitly casting `a` to `float32` before summing fixes the issue. Without the explicit casting, it seems the JIT casts `a` from `int64` to `float64`.
> 
> Knowing this is probably enough to write `numba_dpex` kernels that can work on all devices but one must think of casting explicitly to have compatibility with igpus.

Most likely it's an issue with type inference. Let's tag @diptorupd he might can give us more pointers.

> In fact forget what I just said, this works:
> 
> ```python
> import dpctl.tensor as dpt
> import dpnp
> import numba_dpex as dpx
> import math
> 
> N = 10
> 
> cst = dpnp.float32(0.5)
> 
> @dpx.kernel
> def func(a):
>     i = dpx.get_global_id(0)
>     i = math.ceil(dpnp.float32(i) + cst)
>     a[i] = a[i] + dpnp.float32(i)   # <<<< change is there
> 
> a = dpt.ones(N, dtype=dpnp.float32)
> 
> func[dpx.Range(N)](a)
> 
> print(dpt.asnumpy(a))
> ```

But I have another question, ` i = math.ceil(dpnp.float32(i) + cst)` is supposed to give an indexing error at the last value of `i`. If `a = [0, 1, 2]`, then `i` is `0 <= i <= 2`. At `i = 2`,  `i = math.ceil(dpnp.float32(i) + cst)` should evaluate to `i = math.ceil(2.0 + 0.5) = 3`. But then why the kernel is not complaining about that?

numba_dpex/tests/kernel_tests/test_math_functions.py
```diff
@@ -24,31 +35,44 @@ def unary_op(request):
...
     np_uop = getattr(np, unary_op)
 
     @dpex.kernel
-    def f(a, b):
+    def f(a, b, c):
         i = dpex.get_global_id(0)
-        b[i] = uop(a[i])
+        k = uop(a[i])
+        if type(k) is int or type(k) is np.int32 or type(k) is np.int64:
+            c[i] = 3
```
AlexanderKalistratov, line 65: This test actually doesn't verify if `ceil` and `floor` works properly. It only checks type.
Please:
* Leave original `test_binary_ops` as is
* Create separate test to test type
* Add test for integer data?

Also we must be sure, that `ceil` and `floor` functions actually tested on data with fraction.

numba_dpex/ocl/mathdecl.py
```diff
@@ -230,10 +230,24 @@ class Math_atanh(Math_unary):
...
+        signature(types.uint64, types.uint64),
+        signature(types.int64, types.float32),
+        signature(types.int64, types.float64),
+    ]
 
 
 class Math_ceil(Math_unary):
     key = math.ceil
+    cases = [
+        signature(types.intp, types.intp),
```
AlexanderKalistratov, line 245: This fix is suspicious.
Original `OpenCL` function has `float (float)` signature. But we are telling Numba that it is `int (float)` signature.
If this really works that means, that `llvm` (we are providing `float (float)` siganture to `llvm`) introduce somewhere cast. It is better to either to add a comment why this is working at all, or to reimplement it in Numba adding explicit cast.

Also for `int` input we could do nothing, but instead we would cast `int` to `float`, then call `ceil` and then cast back.

@diptorupd what is your opinion?

AlexanderKalistratov (changes requested):

AlexanderKalistratov:
@fcharras 
> Knowing this is probably enough to write numba_dpex kernels that can work on all devices but one must think of casting explicitly to have compatibility with igpus.

We are aware of this issue and it is going to be addressed in future. Unfortunately I can't provide you any estimations at the moment.

AlexanderKalistratov:
@chudur-budur 
> But I have another question,  i = math.ceil(dpnp.float32(i) + cst) is supposed to give an indexing error at the last value of i. If a = [0, 1, 2], then i is 0 <= i <= 2. At i = 2, i = math.ceil(dpnp.float32(i) + cst) should evaluate to i = math.ceil(2.0 + 0.5) = 3. But then why the kernel is not complaining about that?

`Numba` doesn't do boundary checks as it has devastating effect on performance.

numba_dpex/ocl/mathdecl.py
diptorupd: @AlexanderKalistratov The flow is something like this (I might be missing few things): The `ocl/mathdecl.py` module registers the allowed signatures for the `math.ceil` function that are provided in the `DpexKernelTargetContext`. Then in the `ocl/mathimpl.py` module we register the actual functions that should be used by Numba when lowering a call to `math.ceil`. 

After that when Numba lowers an actual instance of the call, it uses the usual machinery inside `numba/core/cpython/mathimpl`, but looks up the registery and uses our math function than the default one. It is inside Numba that the cast is applied when it sees that a specific `math.ceil` is supposed to return an integer. 

https://github.com/numba/numba/blob/30df74943112acc2706a98630bba6ca1f310e0c8/numba/cpython/mathimpl.py#L108.

diptorupd:
> @fcharras
> 
> > Knowing this is probably enough to write numba_dpex kernels that can work on all devices but one must think of casting explicitly to have compatibility with igpus.
> 
> We are aware of this issue and it is going to be addressed in future. Unfortunately I can't provide you any estimations at the moment.

The solution I have in mind for a while is to split up `DpexKernelTargetContext` that is now used to compile all kernels into two separate contexts: `DpexFP32KernelTargetContext` and `DpexFP64KernelTargetContext`. Once we have two separate contexts we will also have two separate registries for math functions. Then, at the point of compiling we can pick the right context based on what device is targeted.

I have not yet worked out all the details, but it should not be a huge effort to prototype. Also, the approach may not solve all issues around lack of fp64 support. I hope to get it in for the next release.

diptorupd:
The PR is out of date and bitrotten and as stated in https://github.com/IntelPython/numba-dpex/pull/960#issuecomment-1480435881 the approach needs to be better thought out to solve the underlying issue.

Closing the PR.

PR #963, opened 2023-03-07
Proper DpctlSyclQueue support

### LINKED ISSUE #930: Properly support dpctl.SyclQueue as a data type supported inside numba-dpex
The initial support for the `dpctl.SyclQueue` type added in 7fe8d9b85edf0ddf587b35e6a0f696801944ef04 does not allow using a `dpctl.SyclQueue` object inside a `dpjit` function. The basic support only allows passing a queue in and returning it from a `dpjit` function.

To properly support `dpctl.SyclQueue` type, the following tasks (and possibly more) needs to be completed:

1. Add a numba struct type representation for `dpctl.SyclQueue`.
2. Define a struct data model for the numba-dpex type representing a Python `dpctl.SyclQueue`.
3. Add unboxing that can store the `DPCTLSyclQueueRef` pointer from a `dpctl.SyclQueue` object into a Numba native struct.
4. Add boxing function to create a `dpctl.SyclQueue` object from a Numba native struct.
5. Add a `make_queue` function for internal usage, similar to the `make_array` function.

chudur-budur:
Re-implementation of the `DpctlSyclQueue` type in numba-dpex.
    
 - The `DpctlSyclQueue` is no longer a numba `OpaqueType`. It is reimplemented as a `StructType` that stores the native `queue_ref` pointer and the parent `PyObject` pointer.
- Introduce C runtime helper functions for boxing/unboxing a `dpctl.SyclQueue`

Fixes #930 


numba_dpex/core/datamodel/models.py
```diff
@@ -54,6 +59,21 @@ def __init__(self, dmm, fe_type):
         super(ArrayModel, self).__init__(dmm, fe_type, members)
 
 
+class ExecutionQueueModel(ListModel):
```
AlexanderKalistratov, line 62: Why `ListModel`?

chudur-budur: We are not decided yet, just a place holder for now.

numba_dpex/core/datamodel/models.py
```diff
@@ -106,3 +93,29 @@ def _init_data_model_manager():
...
 register_model(DpctlSyclQueue)(OpaqueModel)
 dpex_data_model_manager.register(DpctlSyclQueue, OpaqueModel)
+
+
+@register_model(SyclQueueType)
+class SyclQueueModel(models.StructModel):
+    def __init__(self, dmm, fe_type):
+        members = [
+            ("backend", types.PyObject),
+            ("driver_version", types.StringLiteral),
```
AlexanderKalistratov, line 103: There is literally no point in add literals to model. If you need literals you should add it to type instead.

numba_dpex/core/datamodel/models.py
```diff
@@ -86,3 +93,29 @@ def _init_data_model_manager():
...
+            ("has_enable_profiling", types.BooleanLiteral),
+            ("is_in_order", types.BooleanLiteral),
+            ("name", types.StringLiteral),
+            ("sycl_context", types.PyObject),
+            ("sycl_device", types.PyObject),
```
AlexanderKalistratov, line 108: Are you going to use `sycl_device` inside compiled code somehow?

AlexanderKalistratov:
I don't get the exact purpose of this PR. Does it adds some new functionality?
If so, do we have tests for this functionality?
Or this is just refactoring and functionality is going to be added later?

Anyway, I don't see any tests for implemented code. E.g. `queue_ref` added, but no tests for it.

mingjie-intel (approved):
Looks good.

diptorupd:
> I don't get the exact purpose of this PR. Does it adds some new functionality? If so, do we have tests for this functionality? Or this is just refactoring and functionality is going to be added later?
> 
At this point there are no user-visible changes. It only changes the implementation inside numba-dpex of the `DpctlSyclQueue` type. Instead of defining the type as a `PyObject` we now use a Numba struct type. 

The next set of changes to use the new type inside numba-dpex and avoid extra queue creation will be the next PR. The changes here should also make it simpler to implement the changes in your dpnp overload PR.

> Anyway, I don't see any tests for implemented code. E.g. `queue_ref` added, but no tests for it.
I am working on adding a test case.

diptorupd:
> I am working on adding a test case.

@AlexanderKalistratov I have added a unit test. Could you please review again? The test evaluates if the `queue_ref` stored in the native representation of a `dpctl.SyclQueue` can be properly accessed inside numba-dpex and then passed to a dpctl C function from inside the LLVM module.

Issue #965, opened 2023-03-10
Potential use of undefined pointer in _dpexrt_python.c


diptorupd:
Identified by static analysis

`Null pointer 'miobj' that comes from line 888 may be dereferenced at line 981./home/sat_bot/base/conda-bld/numba-dpex_1678416609610/work/numba_dpex/core/runtime/_dpexrt_python.c:981 |DPEXRT_sycl_usm_ndarray_to_python_acqref()`

PR #972, opened 2023-03-13
Handle the case when arraystruct->meminfo is null to close gh-965

oleksandr-pavlyk:
Closes gh-965 by raising Python exception when `arraystruct->meminfo` field is null, and unhandled exception would have caused dereferencing of the null pointer and subsequent crash.

### DIFF
--- a/numba_dpex/core/runtime/_dpexrt_python.c
+++ b/numba_dpex/core/runtime/_dpexrt_python.c
@@ -959,6 +959,13 @@ DPEXRT_sycl_usm_ndarray_to_python_acqref(arystruct_t *arystruct,
         }
         Py_DECREF(args);
     }
+    else {
+        PyErr_Format(PyExc_ValueError,
+                     "In 'DPEXRT_sycl_usm_ndarray_to_python_acqref', "
+                     "failed to create a new MemInfoObject object since "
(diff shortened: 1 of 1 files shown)


chudur-budur (commented):
LGTM

diptorupd:
@oleksandr-pavlyk @chudur-budur the solution is fine for now especially as it has already been merged. 

I wanted to implement it in a different way. Look at what Numba does for NumPy arrays: https://github.com/numba/numba/blob/ce69f3010c77942123815d86900cb841c0e5d2fa/numba/core/runtime/_nrt_python.c#L366

Numba uses the `arraystruct.meminfo` if the `meminfo` object is populated. Otherwise it lets the `ndarray` parent to be `NULL` when calling `PyArray_NewFromDescr`.

```
    array = (PyArrayObject *) PyArray_NewFromDescr(retty, descr, ndim,
                                                   shape, strides, arystruct->data,
                                                   flags, (PyObject *) miobj);
```
We should have equivalent C-API for tensor and dpnp and handle the case where we do not have `meminfo`.

oleksandr-pavlyk:
@diptorupd I think it should be reimplemented as needed. The fix went in like this to comply with SDL requirements.

PR #973, opened 2023-03-16
Add Dockerfile and workflow

### LINKED ISSUE #933: Providing Docker packages and a Dockerfile for numba-dpex
Providing a Dockerfile that creates an environment with all dependencies will help collaborators and users of `numba-dpex` to quickly setup a working environment to build and test `numba-dpex`. 

The soda-inria group has a Dockerfile (https://github.com/soda-inria/sklearn-numba-dpex/blob/main/docker/Dockerfile) that sets up all needed dependencies from all dpex packages: `dpnp`, `dpctl` and `numba-dpex` and then builds these three packages from source. We can use that Dockerfile as a basis to create a Dockerfile for `numba-dpex`.

The Dockerfile should achieve the following: 

- [ ] Install all needed driver and runtime dependencies
- [ ] Build dpnp and dpctl from source and install them
- [ ] Install the needed llvm-spirv binary from dpcpp (like done in soda-inria/sklearn-numba-dpex)

Once that is done, we should be able to clone numba-dpex and build and run it in the Docker container.

ZzEeKkAa:
Add Dockerfile both for runtime and building with and without gpu drivers

- [x] Have you made sure that new changes do not introduce compiler warnings? (n/a)

Workflow will produce these docker tags:
https://github.com/ZzEeKkAa?tab=packages&repo_name=numba-dpex

Action result can be found here:
https://github.com/ZzEeKkAa/numba-dpex/actions/runs/4473683752

There where few issues that prevented docker from building:
- dpctl was not compiling on Debian based machine (fixed in dpctl repo)
- latest graphics drivers fail numba_dpex test for kernel that tries to print from GPU

Fixes #933

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,456 @@
+# syntax=docker/dockerfile:1.3
+# NB: at least 1.3 is needed to benefit from ARG expansion in bind mount arguments
+ARG PYTHON_VERSION=3.9.16
+
+# Driver args
+# print on gpu is broken for 22.43.24595.30 + igc-1.0.12812.26
+ARG CR_TAG=22.43.24595.30
+ARG IGC_TAG=igc-1.0.12504.5
+ARG CM_TAG=latest
+ARG L0_TAG=latest
+
+# ONEAPI
+ARG ONEAPI_INSTALLER_URL=https://registrationcenter-download.intel.com/akdlm/irc_nas/19079
+ARG ONEAPI_VERSION=2023.0.0
+ARG ONEAPI_INSTALL_BINARY_NAME=l_BaseKit_p_$ONEAPI_VERSION.25537.sh
+ARG ONEAPI_INSTALL_DIR=/opt/intel/oneapi
+
+# Versions of the intel python packages
+ARG DPCTL_GIT_BRANCH=0.14.2
+ARG DPCTL_GIT_URL=https://github.com/IntelPython/dpctl.git
+
+ARG DPNP_GIT_BRANCH=0.11.1
+ARG DPNP_GIT_URL=https://github.com/IntelPython/dpnp.git
+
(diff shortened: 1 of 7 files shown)


docs/user_guides/getting_started.rst
```diff
@@ -103,6 +103,54 @@ To run the examples:
 
     python numba_dpex/examples/sum.py
 
+Docker
+------
+
```
diptorupd, line 108: Need to add instruction/information about the other two packages: `builder`, `stages`.

Dockerfile
```diff
@@ -0,0 +1,456 @@
...
+
+# Driver args
+# print on gpu is broken for 22.43.24595.30 + igc-1.0.12812.26
+ARG CR_TAG=22.43.24595.30
+ARG IGC_TAG=igc-1.0.12504.5
+ARG CM_TAG=latest
+ARG L0_TAG=latest
+
+# ONEAPI
+ARG ONEAPI_INSTALLER_URL=https://registrationcenter-download.intel.com/akdlm/irc_nas/19079
```
diptorupd, line 13: Is this something we have keep changing with every oneAPI release?

diptorupd:
Great work @ZzEeKkAa! 

I lightly tested the runtime package and so far so good :+1: We have to see why the two windows CI run failed, they seem to be unrelated to this PR.

Dockerfile
ZzEeKkAa: I guess so. I'd like to find way how to deal with it, just setting ONEAPI_VERSION, not build number as well

docs/user_guides/getting_started.rst
ZzEeKkAa: Done.

diptorupd (approved):

PR #979, opened 2023-03-22
Replaced llvm_spirv from oneAPI path by dpcpp-llvm-spirv package.

mingjie-intel:


diptorupd:
@mingjie-intel we are using `llvm-spirv` inside the setup.py to pre-compile the atomic emulation. That too should be updated.

PR #992, opened 2023-04-06
Add kernel simulator

Hardcode84:
Add kernel api software simulator.
* Simulator runs completely in interpreted mode, no compilation or type inference is done.
* Simulator consists of 2 parts:
  * `kernel_sim_impl.py` - api-independent kernel implementation
    * provides id/size/allocation/barrier proxies and `execute_kernel` function, which is simulator entry point.
    * Simulator user must provide `need_barrier` flag if kernel need any cross-workitem synchronization (or just always set it `True` in conservative case)
    * User must also provide `replace_global_func` hook to replace frontend-specific APIs with simulator calls (see `_replace_global_func` in  `kernel.py`)
  * `kernel.py` - wrapper for `numba-dpex` kernel api
* Simulator need `greenlet` package to handle barriers and synchronization 
* Not all `numba-dpex` kernel APIs covered yet
* Added simulator-based `get_global/local_id` tests as example


Hardcode84:
@diptorupd it's ready for review

diptorupd:
> @diptorupd it's ready for review

@Hardcode84 few general questions:

a) If we have this simulator should the dpex intrinsic functions like `get_global_id` be implemented as stubs? (refer https://github.com/IntelPython/numba-dpex/blob/main/numba_dpex/ocl/stubs.py). Is a better way to implement the intrinsic functions as overloads of the functions in the `kernel_sim`? Then the usage model can become: a user writes a kernel function in Python (with kernel API restrictions) and once they are ready they add `@kernel` decorator?

b) Another approach may be to add a kwarg to the kernel decorator (may be `simulator=True`) and invoke simulator that way?

What do you think?

Also, please add an example and eventually we will need docs before beta release.

Hardcode84:
> > @diptorupd it's ready for review
> 
> @Hardcode84 I will review in more detail, but starting with few general questions regarding UX:
> 
> a) Once we have the simulator we can re-implement the dpex intrinsic functions like `get_global_id` as overloads of the functions in the `kernel_sim`? Then the usage model can become: a user writes a kernel function in Python (with kernel API restrictions) and once they are ready they add `@kernel` decorator?
> 
> b) Another approach may be to add a kwarg to the kernel decorator (may be `simulator=True`) and invoke simulator that way?
> 
> What do you think?
> 
> Also, please add an example and eventually we will need docs before beta release.

I'm not sure exposing simulator api directly to user is good idea,  but we can call it from our `get_xyz_id` stubs , with proper error reporting if called if outside kernel context.

I initially did simulator as separate decorator to reduce cross-dependencies (simulator depends on actual kernel api but not the other way around).

Regarding separate decorator vs `simulator=True` arg from user perspective I don't have strong preferences either side.

diptorupd:
> Regarding separate decorator vs `simulator=True` arg from user perspective I don't have strong preferences either side.

Ok we can address it as a follow up.

diptorupd:
@Hardcode84 Can you squash some of the commits together? Let us have three commits: one for the main changes, one for the unit test and one for the example.

I think after that it is good to go. We can further improve as follow up work.

diptorupd:
Thank you @Hardcode84!

We took several inspiration from your PR and the new `kernel_api` module in `numba_dpex` is a direct result of that. As the `kernel_api` design is now tailored towards SYCL API rather than OpenCL. Your PR does not directly apply. I am closing this PR for that reason.

PR #993, opened 2023-04-07
Fix typo in the keyword argumnet when constructing a tensor.

diptorupd:
   - numba-dpex's usm_ndarray type relies on dpctl.tensor to determine the default dtype for an array. In the constructor for the usm_ndarray type, there was a typo that was causing an exception in dpjit when dpnp.empty or other constructor was called without a dtype.
   - Adds a unit test case.


mingjie-intel (approved):
Added one check to check shape to be a positional only argument.

PR #1003, opened 2023-04-14
Replace parent package for cached_property import

ZzEeKkAa:
For some reason it is no longer available via `numba.core.utils` import. Anyway, it looks like clean up change.
https://github.com/IntelPython/numba-dpex/issues/1000

~UPD: there are few places to face similar issue.~


diptorupd (approved):
Thanks!

PR #1004, opened 2023-04-14
Fix fill_arrayobj

chudur-budur:
**Some fixes:**  
  - `fill_arrayobj()` takes `sig`, `args` and also `fill_value`, this duplication is not needed, since the value is supposed to be in `args` already.
  - In the `fill_arrayobj()`, we check if the `fill_value` is `float` by probing `args[1]`. But when there is no `fill_value` in `args`, we are testing on a wrong argument, this happens in the case of `dpnp.ones()`, `dpnp.zeros()` etc. which is not correct. *But it doesn't produce any bug or incorrect result, since the `fill_value` is either `0` or `1`.*
  - Now the `codegen()` functions in all intrinsics are cleaner and streamlined.

### DIFF
--- a/numba_dpex/dpnp_iface/_intrinsic.py
+++ b/numba_dpex/dpnp_iface/_intrinsic.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from llvmlite import ir as llvmir
+from llvmlite.ir.types import DoubleType, FloatType
 from numba import types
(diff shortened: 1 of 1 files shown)


diptorupd:
@chudur-budur The PR is good to go as is. However, we should explore how to use kwargs for intrinsics.

diptorupd (approved):
Good improvements!

PR #1002, opened 2023-04-14
Rename nrt_debug_print -> drt_debug_print

ZzEeKkAa:
Numba package updated syntax for `nrt_debug_print`, which caused issues on compilation time. This PR renames function in order to be able to compile for both versions of numba and be more reliable on numba updates.

https://github.com/IntelPython/numba-dpex/issues/1000

- [x] Have you provided a meaningful PR description?
- [x] Have you added a test, reproducer or referred to an issue with a reproducer?
- [x] Have you tested your changes locally for CPU and GPU devices?
- [x] Have you made sure that new changes do not introduce compiler warnings?
- [x] If this PR is a work in progress, are you filing the PR as a draft?

PR #1005, opened 2023-04-16
Move the declaration of drt_debug_print into a header.

diptorupd:
    - Fixes build error after the nrt_dbg_print was renamed to
      drt_debug_print.

### DIFF
--- a/numba_dpex/core/runtime/_dbg_printer.h
+++ b/numba_dpex/core/runtime/_dbg_printer.h
@@ -26,3 +26,15 @@
         X;                                                                     \
     }
 #endif
+
+/*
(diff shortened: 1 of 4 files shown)


diptorupd:
The issue was after #1002 we renamed the function in all files but it was not publicly declared and defined, causing an:

```bash
error: call to undeclared function 'drt_debug_print'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
``` 
The fix is to move the function definition out of `_dpexrt_python.c` into `_dbg_printer.h` and then include `_dbg_printer.h` everywhere.

PR #1007, opened 2023-04-18
Enabling sycl_queue for USMNdArray in dpnp overloads

chudur-budur:
- This is PR enables `dpctl.SyclQueue` for `USMNdArray` in `dpnp` overloads. 
- This PR does not pass SYCL queue to the functions in `numba_depx/core/runtime/_dpexrt_python.c` (yet).
- Since this PR, we won't have `"unknown"` as a default device string anymore, we will use `None` instead, when there is no need to specify the device type.

### DIFF
--- a/numba_dpex/core/types/usm_ndarray_type.py
+++ b/numba_dpex/core/types/usm_ndarray_type.py
@@ -12,6 +12,7 @@ from numba.core.typeinfer import CallConstraint
 from numba.core.types.npytypes import Array
 from numba.np.numpy_support import from_dtype
 
+from numba_dpex.core.types.dpctl_types import DpctlSyclQueue
 from numba_dpex.utils import address_space
 
 
@@ -24,42 +25,44 @@ class USMNdArray(Array):
         layout="C",
         dtype=None,
         usm_type="device",
-        device="unknown",
+        device=None,
         queue=None,
         readonly=False,
         name=None,
         aligned=True,
         addrspace=address_space.GLOBAL,
     ):
-        if not isinstance(device, str):
+        if queue is not None and device is not None:
             raise TypeError(
-                "The device keyword arg should be a str object specifying "
-                "a SYCL filter selector"
-            )
-
-        if not isinstance(queue, dpctl.SyclQueue) and queue is not None:
-            raise TypeError(
-                "The queue keyword arg should be a dpctl.SyclQueue object or None"
+                "numba_dpex.core.types.usm_ndarray_type.USMNdArray.__init__(): "
+                "`device` and `sycl_queue` are exclusive keywords, i.e. use one or other."
             )
 
         self.usm_type = usm_type
         self.addrspace = addrspace
 
-        if device == "unknown":
-            device = None
-
-        if queue is not None and device is not None:
-            raise TypeError(
-                "'queue' and 'device' keywords can not be both specified"
-            )
-
         if queue is not None:
+            if not isinstance(queue, dpctl.SyclQueue):
+                raise TypeError(
+                    "numba_dpex.core.types.usm_ndarray_type.USMNdArray.__init__(): "
+                    "The queue keyword arg should be a dpctl.SyclQueue object or None."
+                )
             self.queue = queue
         else:
             if device is None:
-                device = dpctl.SyclDevice()
+                sycl_device = dpctl.SyclDevice()
+            else:
+                if not isinstance(device, str):
+                    raise TypeError(
+                        "numba_dpex.core.types.usm_ndarray_type.USMNdArray.__init__(): "
+                        "The device keyword arg should be a str object specifying "
+                        "a SYCL filter selector."
+                    )
+                sycl_device = dpctl.SyclDevice(device)
 
-            self.queue = dpctl.get_device_cached_queue(device)
+            self.queue = dpctl._sycl_queue_manager.get_device_cached_queue(
+                sycl_device
+            )
 
         self.device = self.queue.sycl_device.filter_string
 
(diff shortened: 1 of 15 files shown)


numba_dpex/dpnp_iface/arrayobj.py
```diff
@@ -126,19 +126,20 @@ def _parse_device_filter_string(device):
     elif isinstance(device, str):
         return device
     else:
-        raise TypeError(
-            "The parameter 'device' is neither of "
-            + "'str' nor 'types.StringLiteral'"
-        )
+        return None
```
diptorupd, line 129: A `TypeError` should be raised if the device is not a string. Silently returning `None` can hide downstream issues.

numba_dpex/core/types/usm_ndarray_type.py
```diff
@@ -24,42 +24,44 @@ def __init__(
...
-        if device == "unknown":
-            device = None
-
-        if queue is not None and device is not None:
-            raise TypeError(
-                "'queue' and 'device' keywords can not be both specified"
-            )
-
-        if queue is not None:
+        if queue and not isinstance(queue, types.misc.Omitted):
```
diptorupd, line 43: If we have a non-None queue argument can it be anything but a `dpctl.SyclQueue`? Where can we end up with the queue being a `types.misc.Omitted`?

numba_dpex/dpnp_iface/arrayobj.py
```diff
@@ -158,24 +159,29 @@ def build_dpnp_ndarray(
...
             corresponding to a non-partitioned SYCL device, an instance of
             :class:`dpctl.SyclQueue`, or a `Device` object returnedby
-            `dpctl.tensor.usm_array.device`. Default: `"unknwon"`.
-        queue (:class:`dpctl.SyclQueue`, optional): Not supported.
-            Default: `None`.
+            `dpctl.tensor.usm_array.device`. Default: `None`.
+        sycl_queue (:class:`dpctl.SyclQueue`, optional): The SYCL queue to use
+            for output array allocation and copying. sycl_queue and device
+            are exclusive keywords, i.e. use one or another. If both are
+            specified, a TypeError is raised unless both imply the same
```
diptorupd, line 166: The last portion of the docstring is no longer valid. Both queue and device cannot be specified anymore.

numba_dpex/dpnp_iface/arrayobj.py
```diff
@@ -158,24 +159,29 @@ def build_dpnp_ndarray(
...
+        errors.TypingError: If both `device` and `sycl_queue` are provided.
 
     Returns:
         DpnpNdArray: The Numba type to represent an dpnp.ndarray.
             The type has the same structure as USMNdArray used to
             represent dpctl.tensor.usm_ndarray.
     """
-    if queue and not isinstance(queue, types.misc.Omitted):
+
+    if sycl_queue and not isinstance(sycl_queue, types.misc.Omitted) and device:
```
diptorupd, line 180: Can be removed as the check is already inside the `UsmNdArray` constructor.

diptorupd, line 180: Can be removed as the check is already inside the `UsmNdArray` constructor.

diptorupd, line 180: Can be removed as the check is already inside the `UsmNdArray` constructor.

numba_dpex/dpnp_iface/arrayobj.py
diptorupd: The same review comment applies to all other docstrings that have same text.

numba_dpex/dpnp_iface/arrayobj.py
```diff
@@ -500,6 +565,128 @@ def impl(
...
+    fill_value,
+    dtype=None,
+    order="C",
+    like=None,
+    device=None,
+    usm_type=None,
+    sycl_queue=None,
+):
+    """Implementation of an overload to support dpnp.full() inside
+    a jit function.
```
diptorupd, line 580: `s/jit/dpjit/d`

numba_dpex/tests/types/USMNdArray/test_exceptions.py
diptorupd, line 1: Prefer a different name. How about `test_array_creation_errors.py`

numba_dpex/core/types/usm_ndarray_type.py
chudur-budur: > If we have a non-None queue argument can it be anything but a `dpctl.SyclQueue`? 

Then we fail on these tests `numba_dpex/tests/dpjit_tests/parfors/test_builtin_ops.py::test_built_in_operators2`

> Where can we end up with the queue being a `types.misc.Omitted`?

The type of queue becomes `types.misc.Omitted` when `dpnp.epmty()` is being called by dpnp functions internally.

numba_dpex/dpnp_iface/arrayobj.py
chudur-budur: > Can be removed as the check is already inside the `UsmNdArray` constructor.

We can actually remove these checks in all `ol_dpnp_<functions>()` too, we are doing this same check on three levels, which we shouldn't do.

numba_dpex/dpnp_iface/arrayobj.py
chudur-budur: Fixed.

chudur-budur:
Superceded by #1022

chudur-budur:
Superceded by #1022

PR #1022, opened 2023-04-26
Fully enable DpctlSyclQueue handling in USMNdArray creation

chudur-budur:
It's a follow-up PR of #1007. This fully implements the support of `dpctl.SyclQueue()` in `@dpjit` so that

- A SYCL queue doesn't need to be created every time when memory allocation happens.
- A single (or cahced) queue can be passed around.
- Functions like `alloc_empty_arrayobj()` and intrinsics like `_empty_nd_impl()` can handle a queue that has been passed down.
- For the same purpose, the memory allocation c interface function like `DPEXRT_MemInfo_alloc()` in `_dpexrt_python.c` takes queue as an argument.

Testing code:
```python
import dpctl
import dpnp
from numba_dpex import dpjit

@dpjit
def func(q1, q2):
    c = dpnp.empty(1, sycl_queue=q1)
    d = dpnp.zeros(1, sycl_queue=q2)
    return c, d

queue1 = dpctl.SyclQueue()
queue2 = dpctl.SyclQueue()
a, b = func(queue1, queue2)
print(a)
print(b)
```


numba_dpex/_patches.py
```diff
@@ -172,8 +173,71 @@ def _mk_alloc(
...
+    #     pyapi.err_set_string(
+    #         "_patches.make_queue(): PyExc_TypeError",
+    #         "can't unbox dpctl.SyclQueue from PyObject into a Numba "
+    #         "native value. The object maybe of a different type",
+    #     )
+
+    queue_struct = builder.load(queue_struct_ptr)
+    queue_ref = builder.extract_value(queue_struct, 1)
+
+    ret = (queue_ref, queue_address_ptr, pyapi)
```
diptorupd, line 232: Change to Namedtuple

numba_dpex/_patches.py
```diff
@@ -230,22 +294,18 @@ def _empty_nd_impl(context, builder, arrtype, shapes):
         )
 
     if isinstance(arrtype, DpnpNdArray):
+        (queue, queue_ptr, pyapi) = make_queue(context, builder, arrtype)
```
diptorupd, line 297: better to rename to queue_ref to be clear what is the object.

numba_dpex/_patches.py
```diff
@@ -172,8 +173,71 @@ def _mk_alloc(
...
+
+    queue_struct = builder.load(queue_struct_ptr)
+    queue_ref = builder.extract_value(queue_struct, 1)
+
+    ret = (queue_ref, queue_address_ptr, pyapi)
+
+    return ret
+
+
 def _empty_nd_impl(context, builder, arrtype, shapes):
```
diptorupd, line 237: The implementation is getting a bit convoluted as we are trying to maintain API consistency with Numba and monkey patch the Numba function. I suggest, we convert the function to dpex-specific and move it into `dpnp_iface/_intrinsics` and make it specific for our needs. Rather than returning a `queue_ref` LLVM IR object, it is a better design to pass in the object, so that a caller has full control over where the ownership of the LLVM IR instruction.

numba_dpex/_patches.py
```diff
@@ -265,6 +325,7 @@ def _empty_nd_impl(context, builder, arrtype, shapes):
         fnop.get_call_type(context.typing_context, sig.args, {})
         eqfn = context.get_function(fnop, sig)
         meminfo = eqfn(builder, args)
+        pyapi.decref(queue_ptr)
```
diptorupd, line 328: Maybe move to end of the function.

numba_dpex/_patches.py
```diff
@@ -294,40 +355,42 @@ def _empty_nd_impl(context, builder, arrtype, shapes):
...
 
-    return ary
+    ret = (ary, queue) if isinstance(arrtype, DpnpNdArray) else ary
+
+    return ret
 
 
 @overload_classmethod(DpnpNdArray, "_usm_allocate")
-def _ol_array_allocate(cls, allocsize, usm_type, device):
+def _ol_array_allocate(cls, allocsize, usm_type, queue):
```
diptorupd, line 364: suggest renaming to `queue_ref`

numba_dpex/core/runtime/_dpexrt_python.c
```diff
@@ -484,10 +479,10 @@ DPEXRT_MemInfo_alloc(npy_intp size, size_t usm_type, const char *device)
 
     mi->size = size;
     mi->external_allocator = ext_alloca;
-    DPEXRT_DEBUG(drt_debug_print(
-        "DPEXRT-DEBUG: DPEXRT_MemInfo_alloc mi=%p "
-        "external_allocator=%p for usm_type %zu on device %s, %s at %d\n",
-        mi, ext_alloca, usm_type, device, __FILE__, __LINE__));
+    // DPEXRT_DEBUG(drt_debug_print(
```
diptorupd, line 482: Update the debug print.

numba_dpex/core/runtime/_dpexrt_python.c
```diff
@@ -440,33 +443,25 @@ static NRT_MemInfo *NRT_MemInfo_new_from_usmndarray(PyObject *ndarrobj,
...
-    {
-        DPEXRT_DEBUG(
-            drt_debug_print("DPEXRT-ERROR: Could not create a sycl::queue from "
-                            "filter string: %s at %s %d.\n",
-                            device, __FILE__, __LINE__));
-        goto error;
-    }
-
     // Allocate a new NRT_ExternalAllocator
     if (!(ext_alloca = NRT_ExternalAllocator_new_for_usm(qref, usm_type)))
```
diptorupd, line 466: There is a potential problem in the design. The qref pointer that is stored in the `NRT_ExternalAllocator` is linked to the Python `DpctlSyclQueue` object in the `USMNdArray` type. If the `USMNdArray` type object goes out of reference the qref is lost and we will take a segfault. I feel that is what is happening in the `blackscholes_njit` example.

Let me think a bit more about how to remediate.

numba_dpex/core/runtime/_dpexrt_python.c
```diff
@@ -1157,6 +1146,11 @@ static int DPEXRT_sycl_queue_from_python(PyObject *obj,
         goto error;
     }
 
+    DPEXRT_DEBUG(DPCTLSyclDeviceRef device_ref;
```
diptorupd, line 1149: Might be useful to convert into a helper function that may be used other places too.

numba_dpex/core/runtime/context.py
```diff
@@ -20,20 +20,20 @@ def _check_null_result(func):
         @functools.wraps(func)
         def wrap(self, builder, *args, **kwargs):
             memptr = func(self, builder, *args, **kwargs)
-            msg = "USM allocation failed. Check the usm_type and filter "
-            "string values."
+            msg = "USM allocation failed. Check the usm_type and filter string values."
```
diptorupd, line 23: Is the msg still relevant?

numba_dpex/dpnp_iface/_intrinsic.py
```diff
@@ -76,7 +76,7 @@ def fill_arrayobj(context, builder, ary, arrtype, fill_value):
     itemsize = context.get_constant(
         types.intp, get_itemsize(context, arrtype[0])
     )
-    device = context.insert_const_string(builder.module, arrtype[0].device)
+    # queue = context.insert_const_string(builder.module, arrtype[0].device)
```
diptorupd, line 79: remove commented out line.

numba_dpex/_patches.py
chudur-budur: Moved to the end but still gives segfault on `blackscholes_njit.py`

numba_dpex/_patches.py
```diff
@@ -230,22 +304,28 @@ def _empty_nd_impl(context, builder, arrtype, shapes):
         )
 
     if isinstance(arrtype, DpnpNdArray):
+        (queue_ref, queue_ptr, pyapi) = make_queue(context, builder, arrtype)
+        # This might fix the segfault
+        # sycl_queue_val = cgutils.alloca_once(
+        #     builder,
+        #     utils.get_llvm_type(context=context, type=types.voidptr),
```
diptorupd, line 311: The problem here is that the context here is LVVM context and not Numba CpuContext.

numba_dpex/_patches.py
```diff
@@ -230,22 +304,28 @@ def _empty_nd_impl(context, builder, arrtype, shapes):
...
+        # )
+        # fn = DpctlCAPIFnBuilder.get_dpctl_queue_copy(
+        #     builder=builder, context=context
```
diptorupd, line 314: Same here

PR #1023, opened 2023-04-28
Fixes stride calculation when unboxing a dpnp ndarray

### LINKED ISSUE #978: Numba-dpex implementation of Rambo fails validation
[Rambo workload's numba-dpex implementation](https://github.com/adarshyoga/dpbench/blob/main/dpbench/benchmarks/rambo/rambo_numba_dpex_n.py) with dpnp calls fails validation. In my initial analysis, I see that the results produced are different to what the numpy/python implementation produces. The numba-dpex prange implementation passes validation. So this problem seems to be specific to dpnp calls.

How to reproduce:

Follow [instructions ](https://github.com/adarshyoga/dpbench/blob/main/README.md)to setup dpbench
Run Rambo - python -c "import dpbench; dpbench.run_benchmark(\"rambo\")

diptorupd:
The strides were computed incorrectly during unboxing a dpnp.ndarray and that was causing the issue reported in #978. The PR fixes the stride computations and adds unit tests.

Fixes #978


adarshyoga:
Tried this PR on rambo implementation in dpbench. The implementation works as expected after this fix.

diptorupd:
Merging based on @adarshyoga's comments.

Issue #1000, opened 2023-04-14
Numba 0.57 compatibility issues


ZzEeKkAa:
This is a meta Issue to evaluate numba-dpex with the upcoming numba 0.57 release (https://github.com/numba/numba/issues/8841)

Theses are blockers for compilation/import of numba_dpex:

- [x] Numba-dpex numba version checking does not handle tag names with `rc` postfix.
- [x] Numba-dpex has compatibility issue with `nrt_debug_print`
- [x] Numba-dpex relies on  module 'numba.core.utils' has no attribute 'cached_property'. Should be 'functools'
- [x] #1028

Tests not passing, so further actions needed. It may be a problem with missing Docker deps (since I've tested it in docker) or more global problems.

This is not full list. Possible issues may appear after resolving list above

diptorupd:
Completed and numba-dpex now supports numba 0.57

PR #1030, opened 2023-05-05
Feature/support numba057

### LINKED ISSUE #1028: Remove usage of lower_extensions
Numba 0.57 removed the use of `lower_extension` for parfor and converted the parfor passes into regular Numba passes rather than passes that are registered using an extension.

Numba-dpex still uses the old `lower_extension` feature in its lowering passes to register a custom parfor lowering pass. The design needs updating to port numba-dpex to Numba 0.57

diptorupd:

Updates numba-dpex to be compatible with numba 0.57. Summary of changes:

- The in-tree copies of `parfor` module has been removed and numba-dpex now uses the upstream `parfor` module and passes.
- The numba-dpex lowering pass has been removed and parfor lowerer has been updated to handle both lowering when Numba's  `sroa`-like optimization is enabled or disabled.
- The Numba monkey patches were refactored and split out of the `_patches` module and moved into a new `numba_pathces` sub-modules.
- The deprecated `lowerer` module has been removed.
- The `NUMBA_DPEX_FALLBACK_ON_CPU` config is removed as there is no fallback to CPU for parfor nodes generated for dpnp array expressions.
- All the parfor lowering modules are now organized into a `numba_dpex.core.parfors` module.

Fixes #1028, #1000


diptorupd:
parfor is broken due to the `mk_alloc` monkeypatch not getting applied properly.

diptorupd:
> parfor is broken due to the `mk_alloc` monkeypatch not getting applied properly.

Fixed the monkey patches. However, something has changes in the typing and dpnp.ndarray is getting inferred as a `types.Array` and not ` DpnpNdArray`. That is leading to a lowering failure.

diptorupd:
Switching dpex to use Numba's default `ParforLower` and the `NativeParforLowering` pass means we do not disable `sroa optimization` always. Tests need to be added to check if dpex works with both SROA diabled and enabled.

lso, debugging no longer uses dpex's DIBuilder. Need to evaluate.

diptorupd:
@ZzEeKkAa once you test out the docker image we are good to go. The Coverage workflow is failing temporarily due to some issues at coveralls.io's end.

diptorupd:
> The Coverage workflow is failing temporarily due to some issues at coveralls.io's end.
The coveralls.io glitch is gone and removing parfors gives us a 15% coverage boost!

PR #1031, opened 2023-05-07
Fully enable DpctlSyclQueue handling in USMNdArray creation

chudur-budur:
It's a follow-up PR of #1007. This fully implements the support of `dpctl.SyclQueue()` in `@dpjit` so that

- A SYCL queue doesn't need to be created every time when memory allocation happens.
- A single (or cahced) queue can be passed around.
- Functions like `alloc_empty_arrayobj()` and intrinsics like `_empty_nd_impl()` can handle a queue that has been passed down.
- For the same purpose, the memory allocation c interface function like `DPEXRT_MemInfo_alloc()` in `_dpexrt_python.c` takes queue as an argument.

Testing code:
```python
import dpctl
import dpnp
from numba_dpex import dpjit

@dpjit
def func(q1, q2):
    c = dpnp.empty(1, sycl_queue=q1)
    d = dpnp.zeros(1, sycl_queue=q2)
    return c, d

queue1 = dpctl.SyclQueue()
queue2 = dpctl.SyclQueue()
a, b = func(queue1, queue2)
print(a)
print(b)
```
- [x] Have you provided a meaningful PR description?
- [ ] Have you added a test, reproducer or referred to an issue with a reproducer?
- [x] Have you tested your changes locally for CPU and GPU devices?
- [x] Have you made sure that new changes do not introduce compiler warnings?
- [ ] If this PR is a work in progress, are you filing the PR as a draft?

PR #1032, opened 2023-05-08
Feature/support sycl queue in array constructor functions

diptorupd:
Supersedes #1031 , #1022

Enables the `sycl_queue` keyword in the various dpnp array constructors (`empty`, `empty_like`, `full`, `full_like`, `zeros`, `zeros_like`, `ones`, `ones_like`). The passed in dpctl.SyclQueue or the one infered during type inference of the `dpnp.ndarray` is used to allocate memory.

The use of filter string in memory allocation is removed.

Improvements to the existing unit test cases and new ones.


numba_dpex/tests/dpjit_tests/dpnp/test_dpnp_empty.py
```diff
@@ -67,3 +91,19 @@ def func(shape):
...
+    assert c.usm_type == dummy_tensor.usm_type
+    assert c.sycl_device == dummy_tensor.sycl_device
+
+
+def test_dpnp_empty_exceptions():
+    queue = dpctl.SyclQueue()
+
+    @dpjit
+    def func2(a, q):
+        c = dpnp.empty_like(a, sycl_queue=q, device="cpu")
```
chudur-budur, line 103: Why `device="cpu"`?

numba_dpex/tests/dpjit_tests/dpnp/test_dpnp_empty.py
```diff
@@ -38,17 +39,40 @@ def func(shape):
...
+
+    if len(c.shape) == 1:
+        assert c.shape[0] == shape
     else:
-        c.sycl_device.filter_string == dpctl.SyclDevice().filter_string
+        assert c.shape == shape
+
+    assert c.dtype == dtype
+    assert c.usm_type == usm_type
+    assert c.sycl_device.filter_string == queue.sycl_device.filter_string
```
chudur-budur, line 68: Why are we comparing on `filter_string`? Why don't we do just `assert c.sycl_device == queue.sycl_device`?

numba_dpex/tests/dpjit_tests/dpnp/test_dpnp_empty_like.py
```diff
@@ -16,23 +16,20 @@
...
     @dpjit
     def func(a):
         c = dpnp.empty_like(a, dtype=dtype, usm_type=usm_type, device=device)
         return c
 
-    if isinstance(shape, int):
-        NZ = numpy.random.rand(shape)
-    else:
-        NZ = numpy.random.rand(*shape)
+    NZ = dpnp.empty(shape)
```
chudur-budur, line 32: The idea was to test if we can create an "empty" array with the same shape from a "non-empty" array. Now `dpnp.random.random()` works, we can use that.

numba_dpex/tests/dpjit_tests/dpnp/test_dpnp_empty_like.py
```diff
@@ -46,13 +43,89 @@ def func(a):
...
+    shape,
+    usm_type,
+    dtype,
+):
+    @dpjit
+    def func(arr):
+        c = dpnp.empty_like(arr)
+        return c
+
+    arr = dpnp.empty(shape=shape, usm_type=usm_type, dtype=dtype)
```
chudur-budur, line 91: We do not need to create another array to check if the `dtype`s are the same, we can just check if the returned array comes with the `dtype` that we provided in the argument.

numba_dpex/tests/dpjit_tests/dpnp/test_dpnp_empty.py
```diff
@@ -67,3 +91,19 @@ def func(shape):
     dummy_tensor = dpctl.tensor.empty(shape)
```
chudur-budur, line 91: We do not need to create another array to check if the `dtype`s are the same, we can just check if the returned array comes with the `dtype` that we provided in the argument.

chudur-budur (commented):
Let's keep things simpler in these test files. For each function, we do these:
1. call dpnp function with default argument
2. call dpnp function with queue and pytest parameterized arguments
2. call dpnp function with device and pytest parameteried arguments
3. call dpnp with both queue and device (exception check)

chudur-budur (commented):
Let's keep things simpler in these test files. For each function, we do these:
1. call dpnp function with default argument
2. call dpnp function with queue and pytest parameterized arguments
2. call dpnp function with device and pytest parameteried arguments
3. call dpnp with both queue and device (exception check)

numba_dpex/tests/dpjit_tests/dpnp/test_dpnp_empty.py
chudur-budur, line 103: Why are we testing `dpnp.empty_like()` inside `dpnp.empty()` test?

diptorupd: It is just a placeholder. The intent is to test if an exception is raised if both `sycl_queue` and `device` are specified.

numba_dpex/tests/dpjit_tests/dpnp/test_dpnp_empty.py
diptorupd: We can try. The reason is that they device objects will not be the same. It depends on how the `dpctlSyclDevice.__eq__` is implemented.

numba_dpex/tests/dpjit_tests/dpnp/test_dpnp_empty_like.py
diptorupd: I do not think it matters. Let us use `dpnp.ones` if you want a non-empty array. Does `dpnp.random.random` work on all architectures?

numba_dpex/tests/dpjit_tests/dpnp/test_dpnp_empty_like.py
```diff
@@ -46,13 +43,89 @@ def func(a):
...
+        func(arr)
+
+
+@pytest.mark.xfail
+def test_dpnp_empty_like_from_freevar_queue():
+    queue = dpctl.SyclQueue()
+
+    @dpjit
+    def func():
+        c = dpnp.empty_like(10, sycl_queue=queue)
```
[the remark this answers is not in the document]
diptorupd: Hmm. I am not passing a dimension, rather a scalar and the output array should be a zero-dimensional array like it is in NumPy. I see that dpnp does not support such a feature yet, so might as well change the test to dpnp.empty.

However, the test is for a different purpose. I accidentally discovered a bug that we cannot use a SyclQueue that is defined as a free variable inside a dpjit function. Maybe the test can be moved to tests for `DpctlSyclQueue` type.

numba_dpex/tests/dpjit_tests/dpnp/test_dpnp_empty_like.py
```diff
@@ -62,7 +135,7 @@ def func1(a):
         return c
 
     try:
-        func1(numpy.random.rand(5, 5))
+        func1(dpnp.empty((5, 5)))
```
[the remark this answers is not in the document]
diptorupd: Does the dpnp random function work on all devices. Let us use ones or zeros perhaps?

numba_dpex/tests/dpjit_tests/dpnp/test_dpnp_empty.py
diptorupd: typo. Fixed

numba_dpex/core/types/usm_ndarray_type.py
```diff
@@ -66,9 +67,22 @@ def __init__(
         self.device = self.queue.sycl_device.filter_string
 
         if not dtype:
-            dummy_tensor = dpctl.tensor.empty(
-                1, order=layout, usm_type=usm_type, sycl_queue=self.queue
-            )
+            if is_fill_value_float:
```
diptorupd, line 70: The logic does not belong to the `UsmNdArray` type. It should be moved into the dpnp overloads

chudur-budur:
I think it's ready for merge. @diptorupd

PR #1054, opened 2023-05-24
support Python 3.11

### LINKED ISSUE #1052: Add support for python 3.11
Numba supports python 3.11 since 0.57.0. Now, that #1030 is merged we can provide this feature to numba-dpex.

I've faced following issue during build:
```
#18 9.187 gcc -pthread -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -I/usr/local/include/python3.11 -I/usr/local/lib/python3.11/site-packages -I/usr/local/lib/python3.11/site-packages/numpy/core/include -I/usr/local/lib/python3.11/site-packages/dpctl/include -I/usr/local/include/python3.11 -c numba_dpex/core/runtime/_dpexrt_python.c -o build/temp.linux-x86_64-cpython-311/numba_dpex/core/runtime/_dpexrt_python.o
#18 9.328 In file included from numba_dpex/core/runtime/_dpexrt_python.c:20:
#18 9.328 numba_dpex/core/runtime/_nrt_python_helper.h:40:2: error: #error "Python minor version is not supported."
#18 9.328    40 | #error "Python minor version is not supported."
#18 9.328       |  ^~~~~
#18 9.344 error: command '/usr/bin/gcc' failed with exit code 1
```

diptorupd:
Enables building numba_dpex with Python 3.11.

Fixes #1052

### DIFF
--- a/numba_dpex/core/runtime/_nrt_python_helper.c
+++ b/numba_dpex/core/runtime/_nrt_python_helper.c
@@ -183,7 +183,7 @@ PyTypeObject MemInfoType = {
  * act as a reminder to update this struct on Python version update! */
 #if (PY_MAJOR_VERSION == 3)
 #if !((PY_MINOR_VERSION == 8) || (PY_MINOR_VERSION == 9) ||                    \
-      (PY_MINOR_VERSION == 10))
+      (PY_MINOR_VERSION == 10) || (PY_MINOR_VERSION == 11))
 #error "Python minor version is not supported."
 #endif
 #else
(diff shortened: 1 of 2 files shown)


ZzEeKkAa:
I'm not that good in CPP designs, but it looks for me more natural to have this check in header file rather than code file. What do you think?

diptorupd:
> I'm not that good in CPP designs, but it looks for me more natural to have this check in header file rather than code file. What do you think?

I agree in general. In this particular case, I went with what Numba does. The check is meant to guard the struct and Numba puts it inside the source file where the struct is defined. Ideally, the struct definition can be moved to a header file and we can reuse Numba's header.

ZzEeKkAa (approved):

PR #1091, opened 2023-07-12
Add dlpack kernel support

ZzEeKkAa:
Add automatic cast dlpackable objects to usm array


ZzEeKkAa:
Does not make sense now. Should be revisited with experimental kernel.

PR #1093, opened 2023-07-14
Make CI more reliable

ZzEeKkAa:
Update is targeted to provide more reliable workflow that picks up all the latest dev dependencies:
- Use `conda-forge` instead of `pkgs/main`
- Use same workflow for linux and windows
- Combine multiple github pages into one workflow file
- Get rid of separate actions since they are not making workflow files more reliable
- ~~Docs does not have `numba_dpex` dependency now and that's make doc build much faster~~
- Use mamba as package resolver for environment
- Use `conda mambabuild` instead of `conda build` provided by `boa` for faster conda package build
- Remove redundant cleanup for conda environment (for Windows it was taking more than a minute)
- Improve general pipeline logic to check that conda package has all the required dependencies
- Prevent multiple messages to appear for github documentation preview

PR checks:


environment/coverage.yml
```diff
@@ -4,29 +4,23 @@ channels:
...
-    - flake8
-    - black==20.8b1
-    - pytest-cov
-    - pytest-xdist
-    - pexpect
-variables:
-  CHANNELS: -c defaults -c numba -c intel -c numba/label/dev -c dppy/label/dev --override-channels
-  CHANNELS_DEV: -c dppy/label/dev -c defaults -c numba -c intel -c numba/label/dev --override-channels
+  - coverage
+  - pre-commit
```
diptorupd, line 21: Why do we really need the linter packages here?

ZzEeKkAa: It was just a copy from pip to conda package. Removed. Looks like it is redundant here.

diptorupd: :+1:

docs/conf.py
```diff
@@ -16,7 +16,7 @@
 
 import sphinx_rtd_theme
 
-import numba_dpex
+# import numba_dpex
```
diptorupd, line 19: These changes seem to superfluous or at least split out into separate commit.

diptorupd (approved):
Great set of changes. Thank you very much!

Issue #760, opened 2022-07-28
When building the wheel the file atomic_ops.spirv does compile but is not included in the archive


fcharras:
Provided a build environment for `numba_dpex`, when running

`python ./setup.py bdist_wheel`

the archive does not include the file `atomic_ops.spir`, when installing the binary later there will not be support for atomic operations. But one can see that the file `atomic_ops.spir`, despite not being included in the archive, was effectively created.

The source of the issue might come from the `MANIFEST.in` or metadata in `setup,py`, those can be tricky to get right.

fcharras:
The workaround I've found is to run `python setup.py develop` before running `python setup.py bdist_wheel`, it ensures that [spirv compile steps](https://github.com/IntelPython/numba-dpex/blob/main/setup.py#L94) are triggered early enough to ensure that the packaging steps will index the binaries that are created during compilation and add it to the tarball.

diptorupd:
@fcharras More of an FYI at this point. I am working on overhauling the atomics support in numba-dpex (https://github.com/IntelPython/numba-dpex/issues/783). The file you point to was added back when there was no native atomic instruction support in the hardware and the file `atomic_ops.spir` provided an emulation layer (copied from what the DPC++ copier used to do) for atomics ops.

Since then dpex does support generating native atomic instructions. The emulator is still the default and generating native atomics is an opt-in feature at the moment by setting `NUMBA_DPEX_ACTIVATE_ATOMICS_FP_NATIVE` to a non zero value.

In my current work, I am evaluating if emulation is still needed and if we can make generation of native atomics instructions the default.

Issue #1068, opened 2023-06-14
Atomic operations on local memory doesn't work on integrated GPU


AlexanderKalistratov:
Reproducer:
```python
# SPDX-FileCopyrightText: 2020 - 2023 Intel Corporation
#
# SPDX-License-Identifier: Apache-2.0

import dpnp as np

import numba_dpex as ndpx


@ndpx.kernel
def atomic_reduction(a):
    local_a_0 = ndpx.local.array(1, dtype=a.dtype)
    lid = ndpx.get_local_id(0)

    if lid == 0:
        local_a_0[0] = 0

    ndpx.barrier(ndpx.LOCAL_MEM_FENCE)

    idx = ndpx.get_global_id(0)
    ndpx.atomic.add(local_a_0, 0, a[idx])

    ndpx.barrier(ndpx.LOCAL_MEM_FENCE)

    ndpx.atomic.add(a, 0, a[idx])


def main():
    N = 10
    a = np.ones(N, dtype='float32')

    print("Using device ...")
    print(a.device)

    atomic_reduction[ndpx.Range(N)](a)
    print("Reduction sum =", a[0])

    print("Done...")


if __name__ == "__main__":
    main()

```

Result:
```
Using device ...
Device(level_zero:gpu:0)
error: 1: Conflicting SPIR-V versions: 1.4 (input modules 1 through 1) vs 1.0 (input module 2).
Traceback (most recent call last):
  File "/home/akalistr/repo/./atomic_op.py", line 42, in <module>
    main()
  File "/home/akalistr/repo/./atomic_op.py", line 35, in main
    atomic_reduction[ndpx.Range(N)](a)
  File "/home/akalistr/miniconda3/envs/dpbench-dev/lib/python3.10/site-packages/numba_dpex/core/kernel_interface/dispatcher.py", line 455, in __call__
    ) = self._compile_and_cache(
  File "/home/akalistr/miniconda3/envs/dpbench-dev/lib/python3.10/site-packages/numba_dpex/core/kernel_interface/dispatcher.py", line 141, in _compile_and_cache
    kernel.compile(
  File "/home/akalistr/miniconda3/envs/dpbench-dev/lib/python3.10/site-packages/numba_dpex/core/kernel_interface/spirv_kernel.py", line 161, in compile
    self._device_driver_ir_module = spirv_generator.llvm_to_spirv(
[...]

Issue #650, opened 2021-11-29
`spirv-link` fails when `NUMBA_DPPY_DEBUG=1`


PokhodenkoSA:
Scenario: "run dpbench gpairs workload"
```bash
NUMBA_DPPY_DEBUG=1 python run_gpairs.py --test --size 64
```
```python-traceback
...
  File "/localdisk/work/spokhode/numba-dppy/numba_dppy/spirv_generator.py", line 115, in link
    check_call(["spirv-link", *flags, "-o", opath, *binaries])
  File "/localdisk/work/spokhode/numba-dppy/numba_dppy/spirv_generator.py", line 37, in check_call
    return _real_check_call(*args, **kwargs)
  File "/localdisk/work/spokhode/miniconda3/envs/dev/lib/python3.9/subprocess.py", line 373, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['spirv-link', '--allow-partial-linkage', '-o', '/tmp/tmpzkvy2w_v/2-linked-spirv', '/tmp/tmpzkvy2w_v/1-generated-spirv', '/localdisk/work/spokhode/numba-dppy/numba_dppy/ocl/atomics/atomic_ops.spir']' returned non-zero exit status 1.
```

PokhodenkoSA:
@reazulhoque 
Any ideas?

PokhodenkoSA:
It works if native atomics enabled.

ZzEeKkAa:
Fixed in #1103

diptorupd:
Native atomics is enabled by default and emulated atomics code is removed. The problem most likely was the `*.cl` kernel for atomics emulation was not pre-compiled with `-g` when the debug flag was used it was linking to a binary without debug symbols.

Since, emulation is now removed the problem is no longer reproducible.

PR #1103, opened 2023-07-26
Remove atomic emulation

ZzEeKkAa:
Since DPCPP now supports atomics, we don't need to emulate them by ourselves.

Fixes: #760, #650, #868, #1068


numba_dpex/ocl/oclimpl.py
```diff
@@ -286,20 +229,15 @@ def native_atomic_add(context, builder, sig, args):
...
-            device_type == dpctl.device_type.gpu
-            and config.NATIVE_FP_ATOMICS == 1
-        ):
-            return native_atomic_add(context, builder, sig, args)
-        else:
-            # Currently, DPCPP only supports native floating point
-            # atomics for GPUs.
-            return atomic_add(context, builder, sig, args, "add")
-    elif dtype == types.int32 or dtype == types.int64:
+    # TODO: do we need this check, or should we just use native_atomic_add for everything?
```
diptorupd, line 234: The check should check if atomic operations are supported for the type, and it should be based on what SYCL supports: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:atomic-references

ZzEeKkAa: Does something like this work:
```python
def support_atomic(dtype: types.Type) -> bool:
    has_atomic64_support = True # TODO: get from dpctl

    # This check should be the same as described in sycl documentation:
    # https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:atomic-references
    return (dtype == types.int32
    or dtype == types.uint32
    or dtype == types.float32
    or (dtype == types.int64
    or dtype == types.uint64
    or dtype == types.float64)
    and has_atomic64_support)
```

numba_dpex/ocl/oclimpl.py
```diff
@@ -282,27 +225,33 @@ def native_atomic_add(context, builder, sig, args):
...
-            device_type == dpctl.device_type.gpu
-            and config.NATIVE_FP_ATOMICS == 1
-        ):
-            return native_atomic_add(context, builder, sig, args)
-        else:
-            # Currently, DPCPP only supports native floating point
-            # atomics for GPUs.
-            return atomic_add(context, builder, sig, args, "add")
-    elif dtype == types.int32 or dtype == types.int64:
+    dtype, device = sig.args[0].dtype, sig.args[0].device
```
ZzEeKkAa, line 248: @diptorupd do you know better way to get device here?

ZzEeKkAa: I've figured it out

ZzEeKkAa:
This PR should unblock these checks:
https://github.com/IntelPython/dpbench/pull/289

numba_dpex/examples/kernel/atomic_op.py
```diff
@@ -10,12 +10,14 @@
 @ndpx.kernel
 def atomic_reduction(a):
     idx = ndpx.get_global_id(0)
```
diptorupd, line 12: Please add some docstring explaining what the example does. It should be pointed out to users the memory order (relaxed) and the scope (device) of the atomic operation.

numba_dpex/ocl/oclimpl.py
```diff
@@ -282,27 +225,29 @@ def native_atomic_add(context, builder, sig, args):
     return builder.call(fn, fn_args)
 
 
+def support_atomic(dtype: types.Type) -> bool:
+    # This check should be the same as described in sycl documentation:
+    # https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:atomic-references
+    # If atomic is not supported, it will be emulated by the sycl compiler.
+    return (
```
diptorupd, line 232: Fine for now. My follow up changes will make it possible to use `dpctl.SyclDevice.has_aspect_atomic64` to refine the check.

diptorupd (changes requested):
The `run_test.sh` script in the `conda-recipe` directory should be updated as well.

The documentation page
https://github.com/IntelPython/numba-dpex/blob/main/docs/source/user_guide/kernel_programming/atomic-operations.rst should be updated as well.

Please test #1068, #774, #711 as well with this PR.

diptorupd:
@ZzEeKkAa in a subsequent PR can you push changes to remove the dependency on spirv-tools?

ZzEeKkAa:
#650 tested:
```cmd
> set NUMBA_DPPY_DEBUG=1
> dpbench -b gpairs -i numba_dpex_k run --no-skip-expected-failures

================ Benchmark GPairs (gpairs) ========================

================ implementation numba_dpex_k ========================
implementation: numba_dpex_k
framework: numba_dpex
framework version: 0.21.1+61.g7d9bcca0.dirty
input size: 8512
setup time: 219.6217ms (219621700 ns)
warmup time: 962.5928ms (962592800 ns)
teardown time: 101.4μs (101400 ns)
max execution times: 4.226ms (4226000 ns)
min execution times: 2.7134ms (2713400 ns)
median execution times: 3.15995ms (3159950 ns)
repeats: 10
preset: S
validated: Success
```

ZzEeKkAa:
This change is ready to go. I've tested all the mentioned issues. 

These issues simply does not make sense now: #760, #868, #711 does not work, see my latest comment there.

ZzEeKkAa:
> @ZzEeKkAa in a subsequent PR can you push changes to remove the dependency on spirv-tools?

Did it here in a separate commit.

PR #1107, opened 2023-08-02
scikit-build setup.py for numba-dpex

chudur-budur:
We need to switch to scikit-build based setup. 

This will make easy to integrate/compile different extensions and libraries with orthogonal requirements.

JIRA Reference: [SAT-6040](https://jira.devtools.intel.com/browse/SAT-6040)

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -17,7 +17,9 @@ requirements:
     host:
         - python
         - setuptools >=63.*
-        - numba 0.57*
+        - scikit-build
+        - cmake
+        - numba >=0.57*
         - dpctl >=0.14*
         - dpnp >=0.11*
         - dpcpp-llvm-spirv
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -17,8 +17,8 @@ requirements:
     host:
         - python
         - setuptools >=63.*
-        - scikit-build
-        - cmake
+        - scikit-build >=0.15*
+        - cmake >=3.26*
         - numba >=0.57*
         - dpctl >=0.14*
         - dpnp >=0.11*
--- a/environment.yml
+++ b/environment.yml
@@ -16,6 +16,8 @@ dependencies:
   - mkl >=2021.3.0  # for dpnp
   - dpcpp-llvm-spirv
   - packaging
+  - scikit-build
+  - cmake
   - pytest
   - pip
   - pip:
--- a/environment.yml
+++ b/environment.yml
@@ -16,8 +16,8 @@ dependencies:
   - mkl >=2021.3.0  # for dpnp
   - dpcpp-llvm-spirv
   - packaging
-  - scikit-build
-  - cmake
+  - scikit-build >=0.15*
+  - cmake >=3.26*
   - pytest
   - pip
   - pip:
--- a/numba_dpex/core/runtime/CMakeLists.txt
+++ b/numba_dpex/core/runtime/CMakeLists.txt
@@ -1,0 +1,124 @@
+#[=======================================================================[.rst:
+_dpexrt_python
+---------------
+
+A cmake file to compile the ``_dpexrt_python`` Python C extension for
+``numba_dpex``. You can build this component locally in-place by invoking these
+commands:
+
+.. code-block:: cmake
+    ~$ cmake .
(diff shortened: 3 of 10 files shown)


numba_dpex/core/runtime/CMakeLists.txt
```diff
@@ -0,0 +1,130 @@
...
+
+# Update CMAKE_MODULE_PATH
+set(SKBUILD_MODULE_PATH ${SKBUILD_PATH}/resources/cmake)
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${SKBUILD_MODULE_PATH})
+
+# Check CMAKE_MODULE_PATH
+message(STATUS "CMAKE_MODULE_PATH=" "${CMAKE_MODULE_PATH}")
+
+# Add packages
+find_package(PythonInterp 3)
```
ZzEeKkAa, line 59: https://cmake.org/cmake/help/latest/module/FindPythonInterp.html

chudur-budur: It's working now, we just had to add `cmake` along with `scikit-build` in the conda environment.

chudur-budur:
@diptorupd @ZzEeKkAa 

I think it's ready for merge now. Could you please review?

diptorupd (changes requested):
The commits should be squashed into a single one.

Also a separate commit should update the Getting Started documentation and README

chudur-budur:
> The commits should be squashed into a single one.

done

> Also a separate commit should update the Getting Started documentation and README

I can't change the docs on this branch, should I push the updated docs directly into the gh-pages branch?

diptorupd:
> I can't change the docs on this branch, should I push the updated docs directly into the gh-pages branch?

You can create a separate branch on IntelPython/numba-dpex with the documentation changes and then open a PR off that branch. gh-pages stores the generated docs and we should never push anything directly to that branch.

conda-recipe/meta.yaml
```diff
@@ -17,7 +17,9 @@ requirements:
     host:
         - python
         - setuptools >=63.*
-        - numba 0.57*
+        - scikit-build
+        - cmake
```
diptorupd, line 21: Do not we need a version check here for cmake? In the `CMakeLists.txt` we have `
cmake_minimum_required(VERSION 3.21...3.27 FATAL_ERROR)`

environment.yml
```diff
@@ -16,6 +16,8 @@ dependencies:
   - mkl >=2021.3.0  # for dpnp
   - dpcpp-llvm-spirv
   - packaging
+  - scikit-build
+  - cmake
```
diptorupd, line 20: same here

numba_dpex/core/runtime/CMakeLists.txt
```diff
@@ -0,0 +1,124 @@
...
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
+
+    if(RET EQUAL "1")
+        message(FATAL_ERROR "Module \'numba\' not found.")
+    endif()
+endif()
+
+# Get dpctl library path
+if(NOT DEFINED DPCTL_LIBRARY_PATH)
```
diptorupd, line 42: How about using the `FindDpctl` cmake macro? https://github.com/IntelPython/dpctl/blob/master/cmake/FindDpctl.cmake

diptorupd: We are actually using the macro on line 84. Where do we need `DPCTL_LIBRARY_PATH`?

conda-recipe/meta.yaml
chudur-budur: Fixed.

environment.yml
chudur-budur: fixed.

numba_dpex/core/runtime/CMakeLists.txt
chudur-budur: That macro doesn't set `DPCTL_LIBRARY_PATH` so we had to set it manually.

chudur-budur:
> The other comment about using the `FindDpctl` macro can perhaps be side stepped for now.
`FindDpctl` macro doesn't set the `dpctl` library directory or path, so we had to set it manually. We can change that macro in `dpctl` upstream open a PR to their repo if needed.

PR #1108, opened 2023-08-07
Removes spirv-tools as a runtime conda dependency.

diptorupd:
#1103 removed `spirv-tools` as a requirement for numba-dpex. The PR removes `spirv-tools` as a run-time dependency from our `meta.yaml`.


ZzEeKkAa:
Thank you, Ditto! I probably missed those two places.

ZzEeKkAa (approved):

PR #1113, opened 2023-08-18
Do not build for python 3.8 in Github CI

### LINKED ISSUE #1114: Drop support for Python 3.8 in GH action conda-packages.yaml workflow
Since NumPy 1.25 requires Python version >3.8, and with https://github.com/IntelPython/dpctl/pull/1282 and https://github.com/IntelPython/dpctl/issues/1295, it makes sense for us to stop building `numba-dpex` for 3.8.

diptorupd:
Python 3.8 is no longer officially supported.

Fixes #1114

### DIFF
--- a/.github/workflows/conda-package.yml
+++ b/.github/workflows/conda-package.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.8', '3.9', '3.10']
+        python: ['3.9', '3.10']
         os: [ubuntu-latest, windows-latest]
 
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


ZzEeKkAa (changes requested):
versioneer.py, docker.yml, _nrt_python_helper.c, and all other occurs in conda-package.yml should be updated as well.

diptorupd:
> versioneer.py, docker.yml, _nrt_python_helper.c, and all other occurs in conda-package.yml should be updated as well.
> 
> P.S.: we can update conda-package.yml, so it takes python versions just from one place, not setting it 3 or 4 times.

Done. I left `versioneer.py` alone as it is a vendored copy of versioneer.

ZzEeKkAa (approved):
LGTM

PR #1115, opened 2023-08-18
Moved examples out of test

roxx30198:
- Moved the examples folder out from tests folder of numba_dpex.
- There are tests inside tests folder that are currently dependent on examples. Is it required to remove those tests, or move the files required inside the test folder itself?


numba_dpex/tests/debugging/common.py
```diff
@@ -13,7 +13,7 @@
 
 def script_path(script):
     package_path = pathlib.Path(numba_dpex.__file__).parent
-    return str(package_path / "examples/debug" / script)
+    return str(package_path / "../examples/debug" / script)
```
diptorupd, line 16: The change will not work as is. Once you move the examples out of tree, any one who installs the package will not be able to run tests from the package as the example path will be invalid.

I suggest moving whatever source files that the debug tests use from the `examples/debug` into the `tests/debugging` directory.

numba_dpex/tests/test_examples.py
```diff
@@ -11,6 +11,6 @@
 
 def test_examples_available():
     package_path = os.path.dirname(numba_dpex.__file__)
-    examples_path = os.path.join(package_path, "examples")
+    examples_path = os.path.join(package_path, "../examples")
```
diptorupd, line 14: Same issue as the debugging test. Once the tests are out of tree, I suggest we remove this particular unit test completely.

numba_dpex/tests/debugging/common.py
roxx30198: Moved files to tests/debugging and updated path.

numba_dpex/tests/test_examples.py
roxx30198: Removed tests_examples.py file from numba_dpex/tests.

PR #1129, opened 2023-09-12
Switch to numba 0.58

ZzEeKkAa:

Closes https://github.com/IntelPython/numba-dpex/issues/1126


diptorupd:
@ZzEeKkAa I feel we should not fully remove the `patch_mk_alloc`. Since, IDP will not yet support Numba 0.58 we can use upstream function if numba version is 0.58, and for lower versions we should use the patch.

That way if someone uses latest numba-dpex from dppy/label/dev, but everything else from IDP that has numba 0.57, they should still face no issues.

ZzEeKkAa:
> @ZzEeKkAa I feel we should not fully remove the `patch_mk_alloc`. Since, IDP will not yet support Numba 0.58 we can use upstream function if numba version is 0.58, and for lower versions we should use the patch.
> 
> That way if someone uses latest numba-dpex from dppy/label/dev, but everything else from IDP that has numba 0.57, they should still face no issues.

In that case we are signing up to support both numba 0.57 and 0.58. I don't see anything criminal here, just something that we need to be aware.

Updated PR so it works both with numba 0.57 and 0.58

diptorupd (approved):

PR #1027, opened 2023-05-04
[WIP] Use the DpexTargetContext inside dpjit

diptorupd:
- [X] Have you provided a meaningful PR description?

The PR is a draft for the set of changes to ensure that numba_dpex.dpjit uses its own target that is isolated from Numba's
CPUTarget. The goal of the PR is to ensure numba-dpex does not have unintended side effects on Numba, and also free numba-dpex to implement custom dispatch, caching, function registry support.

- [ ] Have you added a test, reproducer or referred to an issue with a reproducer?

- [ ] Have you tested your changes locally for CPU and GPU devices?
As yet minimally.
- [ ] Have you made sure that new changes do not introduce compiler warnings?
- [X] If this PR is a work in progress, are you filing the PR as a draft?'

Fixes #1024

PR #1130, opened 2023-09-13
Numba-mlir integration

Hardcode84:
* Add `use_mlir` option to decorator with proper target options support
* Insert mlir parfor lowering passe if flag is set
* Update some prange tests
* One can use `NUMBA_MLIR_LOG_GPU_RUNTIME_CALLS=1` to check if numba-mlir gpu runtime was actually invoked

### DIFF
--- a/numba_dpex/decorators.py
+++ b/numba_dpex/decorators.py
@@ -155,6 +155,7 @@ def dpjit(*args, **kws):
     kws.update({"nopython": True})
     kws.update({"parallel": True})
     kws.update({"pipeline_class": DpjitCompiler})
+    kws.update({"_target": "dpex"})
 
     # FIXME: When trying to use dpex's target context, overloads do not work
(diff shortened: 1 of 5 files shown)


numba_dpex/decorators.py
```diff
@@ -155,6 +155,7 @@ def dpjit(*args, **kws):
     kws.update({"nopython": True})
     kws.update({"parallel": True})
     kws.update({"pipeline_class": DpjitCompiler})
+    kws.update({"_target": "dpex"})
```
diptorupd, line 158: I ran into a problem with this change previously (see the FIXME note few lines below). The `dpnp` overloads are right now added to the CPUTarget and changing dpjit to use `DpexTarget` makes numba not locate the dpnp overloads.
#1027 needs to be completed first before we can do the change.

diptorupd:
Not being considered at this point, we will revive again if we consider integration in the future.

PR #1134, opened 2023-09-21
Add sycl event

ZzEeKkAa:
Add dpctl.SyclEvent to numba friendly types, so that we can convert kernel dispatcher into @dpjit function.

Closes: #1137

### DIFF
--- a/numba_dpex/core/datamodel/models.py
+++ b/numba_dpex/core/datamodel/models.py
@@ -9,7 +9,13 @@ from numba.core.extending import register_model
 from numba_dpex.core.exceptions import UnreachableError
 from numba_dpex.utils import address_space
 
-from ..types import Array, DpctlSyclQueue, DpnpNdArray, USMNdArray
+from ..types import (
+    Array,
+    DpctlSyclEvent,
+    DpctlSyclQueue,
+    DpnpNdArray,
+    USMNdArray,
+)
 
 
 class GenericPointerModel(PrimitiveModel):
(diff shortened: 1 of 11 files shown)


numba_dpex/core/datamodel/models.py
```diff
@@ -197,6 +230,10 @@ def _init_data_model_manager() -> datamodel.DataModelManager:
     # model manager. The dpex_data_model_manager is used by the DpexKernelTarget
     dmm.register(DpctlSyclQueue, SyclQueueModel)
 
+    # Register the DpctlSyclEvent type to SyclQueueModel in numba_dpex's data
+    # model manager. The dpex_data_model_manager is used by the DpexKernelTarget
+    dmm.register(DpctlSyclEvent, SyclEventModel)
```
diptorupd, line 235: I feel adding the `SyclEventModel` to `dpex_datamode_manager` is not needed as an event object should not be used inside a kernel. Same with `SyclQueueModel`.

numba_dpex/core/runtime/_dpexrt_python.c
```diff
@@ -1269,6 +1273,97 @@ static PyObject *DPEXRT_sycl_queue_to_python(queuestruct_t *queuestruct)
...
+
+    event_struct->parent = obj;
+    event_struct->event_ref = event_ref;
+
+    return 0;
+
+error:
+    // If the check failed then decrement the refcount and return an error
+    // code of -1.
+    // Decref the Pyobject of the array
```
diptorupd, line 1321: The comment needs to be removed as well.

numba_dpex/core/runtime/_dpexrt_python.c
```diff
@@ -1269,6 +1273,97 @@ static PyObject *DPEXRT_sycl_queue_to_python(queuestruct_t *queuestruct)
...
+        PyErr_Format(PyExc_ValueError,
+                     "In 'box_from_eventstruct_parent', "
+                     "failed to create a new dpctl.SyclEvent object.");
+        return NULL;
+    }
+
+    DPEXRT_DEBUG(
+        drt_debug_print("DPEXRT-DEBUG: In DPEXRT_sycl_event_to_python.\n"););
+
+    Py_INCREF(orig_event);
```
diptorupd, line 1362: Can you add a comment for the `Py_INCREF` call to explain why it is called and issues if not done. Will save problems in future.

numba_dpex/core/typing/typeof.py
```diff
@@ -2,15 +2,15 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-from dpctl import SyclQueue
+from dpctl import SyclEvent, SyclQueue
 from dpctl.tensor import usm_ndarray
 from dpnp import ndarray
-from numba.extending import typeof_impl
+from numba.extending import get_cython_function_address, typeof_impl
```
diptorupd, line 8: Why do we need to import `get_cython_function_address`?

diptorupd (changes requested):
Approved with some minor changes.

numba_dpex/core/runtime/_dpexrt_python.c
ZzEeKkAa: Let's reconsider if we still need this inc/dec ref. Still the case when it used asynchronously from the python flow may cause gc to deallocate object. It is not a case with numpy in numba because we are doing incref for memory data.

ZzEeKkAa: We need to check if numba does it for us out of the box.

numba_dpex/core/datamodel/models.py
ZzEeKkAa: I'll remove SyclEvent in this PR and SyclQueue in separate PR that will be targeted to fix it along with all other issues.

diptorupd (approved):

PR #1144, opened 2023-09-25
Implement support for dpnp.arange()

chudur-budur:
WIP for implementing overloads of `dpnp.arange()`.
This PR also implements dispatch vector to store template instantiated sycl kernels for basic tensor ops.


chudur-budur:
Deleted by accident.

.clang-format
```diff
@@ -5,6 +5,7 @@ AlignEscapedNewlines: Right
 AllowAllParametersOfDeclarationOnNextLine: false
 BinPackParameters: false
 BreakBeforeBraces: Custom
+SortIncludes: false
```
diptorupd, line 8: Why do we need this change? In general I do not like these type of changes as it indicates that headers should be in a specific order and any inadvertent change will introduce issues.

numba_dpex/core/runtime/kernels/tensor/include/typeutils.hpp
```diff
@@ -0,0 +1,174 @@
...
+#ifndef __TYPES_HPP__
+#define __TYPES_HPP__
+
+#include <cstdlib>
+#include <complex>
+#include <exception>
+#include <utility>
+#include <string>
+#include <cstdint>
+// #include <cxxabi.h> // this is gcc specific, not supported on windows
```
diptorupd, line 14: leftover cleanup?

chudur-budur: This is needed for the `demangle()` function, which is very useful for debugging this kind of templated code.

numba_dpex/core/runtime/context.py
```diff
@@ -21,7 +21,9 @@ def _check_null_result(func):
         @functools.wraps(func)
         def wrap(self, builder, *args, **kwargs):
             memptr = func(self, builder, *args, **kwargs)
-            msg = "USM allocation failed. Check the usm_type and queue."
+            msg = "In {0:s}.wrap(). USM allocation failed at {1:s}(). Check the usm_type and queue.".format(
```
diptorupd, line 24: Please use f-string and look out for 80 char libe width .

numba_dpex/core/runtime/kernels/tensor/include/typeutils.hpp
diptorupd: We should not have commented out code in our sources. We recently added a flake8 check to disallow it explicitly for the Python code.

.clang-format
chudur-budur: Removed.

numba_dpex/core/runtime/kernels/tensor/include/typeutils.hpp
chudur-budur: Fixed.

numba_dpex/core/runtime/context.py
chudur-budur: Fixed.

diptorupd:
Thank you for the exploratory work @chudur-budur!

As discussed, we will in future explore adding support for these constructors using the `overload` functionality and using the kernel api. As such the current PR is out of scope and I am closing it.

PR #1148, opened 2023-09-28
Adds Range and NdRange as supported types in numba_dpex.dpjit.

diptorupd:
    - The numba_dpex.core.kernel_interface.indexers.Range and
      NdRange classes can now be used inside a numba_dpex.dpjit
      decorated function, as arguments to a dpjit decorated function,
      and returned from a dpjit decorated function.
    - Defines a datamodel, type class, overloads for the constructors,
      box, and unbox functions for both classes.
~~The PR should be merged post 0.21.3 tag~~

### DIFF
--- a/numba_dpex/core/datamodel/models.py
+++ b/numba_dpex/core/datamodel/models.py
@@ -14,6 +14,8 @@ from ..types import (
     DpctlSyclEvent,
     DpctlSyclQueue,
     DpnpNdArray,
+    NdRangeType,
+    RangeType,
     USMNdArray,
 )
 
@@ -195,6 +197,39 @@ class SyclEventModel(StructModel):
         super(SyclEventModel, self).__init__(dmm, fe_type, members)
 
 
+class RangeModel(StructModel):
+    """The native data model for a
+    numba_dpex.core.kernel_interface.indexers.Range PyObject.
+    """
+
+    def __init__(self, dmm, fe_type):
+        members = [
+            ("ndim", types.int64),
+            ("dim0", types.int64),
+            ("dim1", types.int64),
+            ("dim2", types.int64),
+        ]
+        super(RangeModel, self).__init__(dmm, fe_type, members)
+
+
+class NdRangeModel(StructModel):
+    """The native data model for a
+    numba_dpex.core.kernel_interface.indexers.NdRange PyObject.
+    """
+
+    def __init__(self, dmm, fe_type):
+        members = [
+            ("ndim", types.int64),
+            ("gdim0", types.int64),
+            ("gdim1", types.int64),
+            ("gdim2", types.int64),
+            ("ldim0", types.int64),
+            ("ldim1", types.int64),
+            ("ldim2", types.int64),
+        ]
+        super(NdRangeModel, self).__init__(dmm, fe_type, members)
+
+
 def _init_data_model_manager() -> datamodel.DataModelManager:
     """Initializes a DpexKernelTarget-specific data model manager.
 
@@ -249,3 +284,8 @@ register_model(DpctlSyclQueue)(SyclQueueModel)
 
 # Register the DpctlSyclEvent type
 register_model(DpctlSyclEvent)(SyclEventModel)
+# Register the RangeType type
+register_model(RangeType)(RangeModel)
+
+# Register the NdRangeType type
+register_model(NdRangeType)(NdRangeModel)
--- a/numba_dpex/core/kernel_interface/indexers.py
+++ b/numba_dpex/core/kernel_interface/indexers.py
@@ -4,6 +4,11 @@
 
 from collections.abc import Iterable
 
+from llvmlite import ir as llvmir
+from numba.core import cgutils, errors, types
+from numba.core.datamodel import default_manager
+from numba.extending import intrinsic, overload
+
 
 class Range(tuple):
     """A data structure to encapsulate a single kernel launch parameter.
@@ -73,6 +78,50 @@ class Range(tuple):
             return self[0] * self[1]
         else:
             return self[0]
+
+    @property
+    def ndim(self) -> int:
+        """Returns the rank of a Range object.
+
+        Returns:
+            int: Number of dimensions in the Range object
+        """
+        return len(self)
+
+    @property
+    def dim0(self) -> int:
+        """Return the extent of the first dimension for the Range object.
+
+        Returns:
+            int: Extent of first dimension for the Range object
+        """
+        return self[0]
+
+    @property
+    def dim1(self) -> int:
+        """Return the extent of the second dimension for the Range object.
+
+        Returns:
+            int: Extent of second dimension for the Range object or -1 for 1D
+            Range
+        """
+        try:
+            return self[1]
+        except:
+            return -1
+
+    @property
+    def dim2(self) -> int:
+        """Return the extent of the second dimension for the Range object.
+
+        Returns:
+            int: Extent of second dimension for the Range object or -1 for 1D or
+            2D Range
+        """
+        try:
+            return self[2]
+        except:
+            return -1
 
 
 class NdRange:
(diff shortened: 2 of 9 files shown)


numba_dpex/core/datamodel/models.py
```diff
@@ -213,3 +253,9 @@ def _init_data_model_manager() -> datamodel.DataModelManager:
 
 # Register the DpctlSyclQueue type
 register_model(DpctlSyclQueue)(SyclQueueModel)
+
+# Register the RangeType type
+register_model(RangeType)(RangeModel)
```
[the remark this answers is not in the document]
diptorupd: No, because these types are not to be used inside kernel. At least that I do not anticipate such a usage right now.

numba_dpex/tests/core/types/range_types/test_unbox_box.py
```diff
@@ -0,0 +1,44 @@
...
+@pytest.mark.parametrize("r", ranges)
+def test_range_unbox_box(r):
+    @dpjit
+    def _tester(r):
+        return r
+
+    r_in = Range(*r)
+    r_out = _tester(r_in)
+
+    assert r_out.ndim == r_in.ndim
```
ZzEeKkAa, line 21: May be we should compare objects and overwrite `_eq` if needed?
```suggestion
    assert r_in == r_out
```

diptorupd: Done in [084cd8b](https://github.com/IntelPython/numba-dpex/pull/1148/commits/084cd8b4785030c2a177b5d292ed8b05a1dc887a)

numba_dpex/core/kernel_interface/indexers.py
```diff
@@ -219,6 +219,12 @@ def __repr__(self):
         """
         return self.__str__()
 
+    def __eq__(self, other):
```
ZzEeKkAa, line 222: why do we have to overload `NdRange` and not overloading `Range`?

diptorupd: `Range` is a sub-type of `tuple` and that is the reason `__eq__` works out of the box.

numba_dpex/core/types/range_types.py
```diff
@@ -0,0 +1,267 @@
...
+        elif typ.ndim == 2:
+            res = c.pyapi.call_function_objargs(class_obj, (dim0_obj, dim1_obj))
+        elif typ.ndim == 3:
+            res = c.pyapi.call_function_objargs(
+                class_obj, (dim0_obj, dim1_obj, dim2_obj)
+            )
+        else:
+            raise ValueError("Cannot unbox Range instance.")
+
+        c.pyapi.decref(ndim_obj)
```
ZzEeKkAa, line 206: Same question as in sycl event case. Do we need to decref here? They are still part of the return object, so they are being referenced. I may miss some part when the reference it increased one more time. Comment with clarification why do we need to do it will be really helpful with understanding.

numba_dpex/core/types/range_types.py
```diff
@@ -0,0 +1,267 @@
...
+        with cgutils.early_exit_if_null(c.builder, stack, dim0_obj):
+            c.builder.store(fail_obj, ret_ptr)
+        dim1_obj = c.box(types.int64, range_struct.dim1)
+        with cgutils.early_exit_if_null(c.builder, stack, dim1_obj):
+            c.builder.store(fail_obj, ret_ptr)
+        dim2_obj = c.box(types.int64, range_struct.dim2)
+        with cgutils.early_exit_if_null(c.builder, stack, dim2_obj):
+            c.builder.store(fail_obj, ret_ptr)
+
+        class_obj = c.pyapi.unserialize(c.pyapi.serialize_object(Range))
```
ZzEeKkAa, line 185: Sorry, lack of knowledge, could you explain why do we do this?

numba_dpex/core/types/range_types.py
```diff
@@ -0,0 +1,267 @@
...
+        ndrange_struct._getvalue(), is_error=c.builder.load(is_error_ptr)
+    )
+
+
+@box(RangeType)
+def box_range(typ, val, c):
+    """
+    Convert a native range structure to a Range object.
+    """
+    ret_ptr = cgutils.alloca_once(c.builder, c.pyapi.pyobj)
```
ZzEeKkAa, line 164: Am I right that this one creates reference on the created object, that's why we don't have to increase reference manually?

ZzEeKkAa (commented):
Overall looks good. I was trying to understand what's going on here, so most of the comments is more to get better understanding, that request to changes something. Also NDRange concept still unclear for me.

numba_dpex/core/types/range_types.py
diptorupd: `ret_ptr` is the `PyObject*` into which the `res` (`Range`) object to be stored is stored. Since it is a new object that gets returned from Numba to CPython, a incref is not needed.

numba_dpex/core/types/range_types.py
diptorupd: Your question actually found a small issue. There is no need to box the `ndim` object at all as it is not stored as an attribute of the `Range` type.

However, your question does apply to the other objects: `dim0_obj`, `dim1_obj`, _etc_.

My reasoning is as follows:

The `dim0_obj` is a new `PyObject` that is returned by the call to the `box` method for that type. The type in this case is an `int`, which is a basic immutable type. Thus, when we pass the `dim0_obj` to the constructor call `c.pyapi.call_function_objargs(class_obj, (dim0_obj,))` the PyObject is passed by "value", _i.e.,_ a new `PyObject` that is now owned by the newly created `Range` instance. The original `dim0_obj` therefore still needs to be properly cleaned up and the `decref` ensures it is eventually cleaned up.

P.S. : Full disclosure, your question forced me to reason about what I was doing, so thank you for it. Otherwise, my code was mostly "monkey see, monkey do" from the Numba extension guide.

numba_dpex/core/types/range_types.py
diptorupd: First a _mea culpa_, I have no idea! Purely "monkey see, monkey do"  from the Numba extension guide.

But, I dug up what it does. Here is the paper trail:

https://github.com/numba/numba/blob/9f74128def1ca51deebfadecc663f0b80f641c57/numba/core/pythonapi.py#L1349

https://github.com/numba/numba/blob/9f74128def1ca51deebfadecc663f0b80f641c57/numba/core/pythonapi.py#L1402

https://github.com/numba/numba/blob/9f74128def1ca51deebfadecc663f0b80f641c57/numba/_helperlib.c#L1058

https://docs.python.org/3/c-api/bytes.html#c.PyBytes_FromStringAndSize

My superficial reading of what is happening here is that for every registered type Numba is maintaining a cache of how to serialize and unserialize an object of that type using the Python `pickle` module. It is a functionality I believe was designed mainly for  on-disk caching of compiled functions and that involves being able to box and unbox argument types. So, we are using that API to call the CPython `PyBytes_FromStringAndSize` to get the `PyObject` structure that represents a `Range` object in this case. Once we have that we are next doing a `PyObject_CallFunctionObjArgs` to call the constructor for that object.

We might need a better CPython expert like @oleksandr-pavlyk for further explanation.

numba_dpex/core/kernel_interface/indexers.py
```diff
@@ -169,3 +218,163 @@ def __repr__(self):
...
+    def codegen(context, builder, sig, args):
+        typ = sig.return_type
+        dim0, dim1, dim2, _ = args
+        range_struct = cgutils.create_struct_proxy(typ)(context, builder)
+        range_struct.dim0 = dim0
+
+        if not isinstance(sig.args[1], types.NoneType):
+            range_struct.dim1 = dim1
+        else:
+            range_struct.dim1 = llvmir.Constant(llvmir.types.IntType(64), -1)
```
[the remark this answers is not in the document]
diptorupd: Addressed in 38ee434c570ed5c70b543fe5e153fe7621521bb1

numba_dpex/core/types/range_types.py
diptorupd: > Your question actually found a small issue. There is no need to box the `ndim` object at all as it is not stored as an attribute of the `Range` type.

Addressed in 38ee434c570ed5c70b543fe5e153fe7621521bb1

numba_dpex/core/kernel_interface/indexers.py
```diff
@@ -74,6 +81,50 @@ def size(self):
...
+    def dim1(self) -> int:
+        """Return the extent of the second dimension for the Range object.
+
+        Returns:
+            int: Extent of second dimension for the Range object or -1 for 1D
+            Range
+        """
+        try:
+            return self[1]
+        except:
```
oleksandr-pavlyk, line 112:
```suggestion
        except IndexError:
```

numba_dpex/core/kernel_interface/indexers.py
```diff
@@ -74,6 +81,50 @@ def size(self):
...
+    def dim2(self) -> int:
+        """Return the extent of the second dimension for the Range object.
+
+        Returns:
+            int: Extent of second dimension for the Range object or -1 for 1D or
+            2D Range
+        """
+        try:
+            return self[2]
+        except:
```
oleksandr-pavlyk, line 125:
```suggestion
        except IndexError:
```

numba_dpex/core/kernel_interface/indexers.py
```diff
@@ -169,3 +220,167 @@ def __repr__(self):
             str: str representation for NdRange class.
         """
         return self.__str__()
+
+    def __eq__(self, other):
+        return (
+            self.global_range == other.global_range
+            and self.local_range == other.local_range
+        )
```
oleksandr-pavlyk, line 228: The `other` can be any type, it should be checked to be of type `NdRange` before attempting to access attributes. Otherwise `AttributeError` would be raised.

numba_dpex/core/kernel_interface/indexers.py
diptorupd: Addressed in af0208dde9d6763117de8789a5924243d60e910e

numba_dpex/core/kernel_interface/indexers.py
diptorupd: Addressed in af0208dde9d6763117de8789a5924243d60e910e

numba_dpex/core/kernel_interface/indexers.py
diptorupd: Addressed in af0208dde9d6763117de8789a5924243d60e910e

diptorupd:
> Also NDRange concept still unclear for me.

https://enccs.github.io/sycl-workshop/expressing-parallelism-nd-range/

numba_dpex/core/types/range_types.py
ZzEeKkAa: Thank you!

numba_dpex/core/types/range_types.py
diptorupd: When a PyObject is created the ref count should be `1`. Once Numba returns the object to CPython, then CPython will track the lifetime of the object, and decref once the object is out of scope and dead. The garbage collector will free it up at that point.

PR #1153, opened 2023-10-02
Improve box unbox tests

ZzEeKkAa:
Updates typing tests to have unified set of tests to run on dpex provided types. Makes sure that ref count is expected. This changes reviled some bugs in reference count manipulations, that were fixed.

- queue inc/ref was not properly matched. Current change fixes it but may result in SyclQueue being garbage collected during run. It is not big deal for now, since all queues being passed through cache that prevents it from being garbage collected.

Checks


numba_dpex/tests/core/types/test_box_unbox.py
```diff
@@ -0,0 +1,145 @@
...
+def constructor_np(request):
+    return request.param
+
+
+@dpx.dpjit
+def unbox_box(a):
+    return a
+
+
+# @numba.njit
```
diptorupd, line 59: please remove commented out code. Also `dpjit` is for now equivalent to `njit(Parallel=True)`

diptorupd (approved):

numba_dpex/tests/core/types/test_box_unbox.py
ZzEeKkAa: Ops, did not see the comment. Let me address this issue in separate PR.

PR #1155, opened 2023-10-03
Fixes bug in boxing a DpnpNdArray from parent.

diptorupd:
    - When boxing a dpnp.ndarray using the reference of the parent
      stored during unboxing, there is a validation step on the strides.
      However, as a dpnp.ndarray object does not store any stride
      information when an array is unit strided the strides need to be
      calculated and validated against the strides in the Numba
      usmarraystruct object.

      The PR fixes the stride calculation that was previously done
      incorrectly.


ZzEeKkAa (approved):
LGTM

Issue #1152, opened 2023-10-02
Having a lot of segfault / incorrect outputs with latest updates


fcharras:
I'm attempting bumps of all the numba_dpex + dpctl stack but a lot of tests at [sklearn-numba-dpex](https://github.com/soda-inria/sklearn-numba-dpex) either segfaults or return wrong results.

Yet again minimal reproducers are tricky to extract, but the regressions are common enough to impact about 10% of the test suite, it's not that rare.

Using GPU devices sometimes lead to segfaults, if using level zero backend it prints:

```
terminate called after throwing an instance of 'sycl::_V1::runtime_error'
  what():  Native API failed. Native API returns: -1 (PI_ERROR_DEVICE_NOT_FOUND) -1 (PI_ERROR_DEVICE_NOT_FOUND)

Aborted (core dumped)
```

with opencl it's a slightly different error:

```
terminate called after throwing an instance of 'sycl::_V1::runtime_error'
  what():  Native API failed. Native API returns: -5 (PI_ERROR_OUT_OF_RESOURCES) -5 (PI_ERROR_OUT_OF_RESOURCES)
Aborted (core dumped)
```

but in both cases it didn't let me find out what is the actual issue. 

When running on CPU, I don't see segfaults but sometimes the tests just fail with wrong outputs.

@diptorupd @oleksandr-pavlyk it would greatly help if you have some insight on things I could try to overcome this. Last time it happened, reverting bumps on drivers solved most of the issues, this time it doesn't seem to be related.

I can reproduce the issue both locally in an iGPU in our custom docker build, but also with the conda install on intel dev cloud with max series gpu.

oleksandr-pavlyk:
@fcharras Try running with {{SYCL_PI_TRACE=-1}} to identify the actual underlying operation that returns the error status leading to terminate. I suspect it is a kernel launch, but it is best confirmed. 

The out of resources hints that the kernel may be asking for too much SLM for example.

fcharras:
(showing the last printed events before terminate) with `level_zero`:

```
---> piEventGetInfo(
	pi_event : 0x55820996a300
	<unknown> : 4563
	<unknown> : 4
	<unknown> : 0x7ffcc61a43ac
	<nullptr>
) ---> 	pi_result : PI_SUCCESS

---> piEventsWait(
	<unknown> : 1
	pi_event * : 0x7f2904000c40[ 0x55820996a300 ... ]
) ---> 	pi_result : -1
[...]

PR #1158, opened 2023-10-04
Add DPEX_OPT, INLINE_THRESHOLD config and do not use numba's OPT

ZzEeKkAa:
It appears that level 3 optimization breaks kernel execution, so we need to set default value to 2 (which seems to work acceptable). This is a quick fix for #1152 and further investigation needed.

### DIFF
--- a/numba_dpex/config.py
+++ b/numba_dpex/config.py
@@ -87,3 +87,7 @@ TESTING_LOG_DEBUGGING = _readenv("NUMBA_DPEX_TESTING_LOG_DEBUGGING", int, DEBUG)
 # Flag to turn on the ConstantSizeStaticLocalMemoryPass in the kernel pipeline.
 # The pass is turned off by default.
 STATIC_LOCAL_MEM_PASS = _readenv("NUMBA_DPEX_STATIC_LOCAL_MEM_PASS", int, 0)
+
+DPEX_OPT = _readenv("NUMBA_DPEX_OPT", int, 2)
+
+INLINE_THRESHOLD = _readenv("NUMBA_DPEX_INLINE_THRESHOLD", int, None)
--- a/numba_dpex/core/codegen.py
+++ b/numba_dpex/core/codegen.py
@@ -2,6 +2,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import logging
+
 from llvmlite import binding as ll
 from llvmlite import ir as llvmir
 from numba.core import utils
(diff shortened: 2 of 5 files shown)


numba_dpex/core/codegen.py
```diff
@@ -31,8 +31,8 @@ def _optimize_final_module(self):
         # Run some lightweight optimization to simplify the module.
         pmb = ll.PassManagerBuilder()
 
-        # Make optimization level depending on config.OPT variable
-        pmb.opt_level = config.OPT
+        # Make optimization level depending on config.DPEX_OPT variable
+        pmb.opt_level = config.DPEX_OPT
```
diptorupd, line 35: We have to check here if the DPEX_OPT was set to `3` and then reset it to `2` and give the user a warning

ZzEeKkAa: If it is by default 2 we can just warn user that it may be dangerous and lead to crashes. I'm thinking of example when user is hacking dpex and really want to see the output...

numba_dpex/core/codegen.py
```diff
@@ -31,8 +31,8 @@ def _optimize_final_module(self):
...
 
         pmb.disable_unit_at_a_time = False
         pmb.inlining_threshold = 2
```
fcharras, line 38: TY for investigating those issues. Also while investigating for https://github.com/IntelPython/numba-dpex/issues/1152 we've also found that `        pmb.inlining_threshold = 2` causes miscompilations. Could it be disabled by default ?

ZzEeKkAa: FYI, opt3 works fine one devcloud with threshold being removed and fails on opt2 with theshold. However I have an environment that fails with opt3 without threshold.

numba_dpex/config.py
```diff
@@ -87,3 +87,19 @@ def __getattr__(name):
 # Flag to turn on the ConstantSizeStaticLocalMemoryPass in the kernel pipeline.
 # The pass is turned off by default.
 STATIC_LOCAL_MEM_PASS = _readenv("NUMBA_DPEX_STATIC_LOCAL_MEM_PASS", int, 0)
+
+DPEX_OPT = _readenv("NUMBA_DPEX_OPT", int, 2)
+
+if DPEX_OPT > 2:
```
diptorupd, line 93: We should not put the check here. Rather I prefer it in the codegen function. The reason is someone can set `config.DPEX_OPT=3` somewhere else in the code without this warning getting triggered.

diptorupd (approved):
Thanks @ZzEeKkAa

PR #1167, opened 2023-10-12
Upload wheel to the same version

ZzEeKkAa:
Wheels are uploaded to the version from wheels package that comes from versioner. That version is different for wheel and Conda and contains GitHub stash. That results into junk in Conda registry version list. PR simply uploads same wheel to the version of Conda package.


oleksandr-pavlyk (approved):
LGTM. Similar change to this was already merged in dpctl: https://github.com/IntelPython/dpctl/pull/1435 and main trunk build is clean.

PR #1178, opened 2023-10-18
An experimental kernel dispatcher that generates a native LLVM wrapper

diptorupd:

Numba-dpex uses the `numba_dpex.core.kernel_interface.JitKernel` class a custom dispatcher class for running kernel functions. The `JitKernel` dispatcher has a different API than `numba.dispatcher.Dispatcher` making it hard to use `JitKernel` as a target dispatcher inside Numba. The limitation means we currently cannot do the following:

- add overloads in the `DpexKernelTarget`
- call kernels from `numba_dpex.dpjit`
- use Numba's unbox/box APIs (numba-dpex uses ctypes to implement the functionality)

There are also potential performance overheads as kernel dispatching fully happens in Python.

The PR introduces a new `KernelDispatcher` class that is a sub-class of `numba.core.dispatcher.Dispatcher`. The `KernelDispatcher` is for now included as an experimental feature, as to be useful all existing `lower` functions in `numba_dpex.ocl.oclimpl` would have to be added to `DpexKernelTargetContext`. The overloads will be done as a follow up PR.

The experimental `KernelDispacther` compiles a `@kernel` decorated function in the following way:

```python
@kernel
def add(a,b,c):
    i = dpex.get_global_id(0)
    c[i] = a[i] + b[i]
```
1. Compile `add` to a LLVM function with the `spir_kernel` calling convention that has a signature: ` void add (DpnpNdarray a, DpnpNdArray b, DpnpNdArray c)`
2. Compiles the LLVM function generated in (1) to SPIR-V using the `dpccp_llvm_spirv` package.
3. Generate a `dpex_kernel_launcher` LLVM module that the following content:
    - The SPIR-V binary generated in (2) as a constant byte array
    - A host callable "launcher" function that has the signature: 
        - ` void add_launcher (Range r, DpnpNdarray a, DpnpNdArray b, DpnpNdArray c)`, or 
        - ` void add_launcher (NdRange ndr, DpnpNdarray a, DpnpNdArray b, DpnpNdArray c)`
    - The "launcher" function does: 
        - create an array to store the kernel arguments, 
        - create an array to store the type ids for kernel arguments, 
        - compile the SPIR-V binary to a `sycl::kernel_bundle`.
        - Extract the `sycl::kernel` from the `sycl::kernel_bundle`
        - Submit the kernel to the `sycl::queue` extracted from a DpnpNdArray kernel argument
        - Wait on the queue
    - A Numba-generated CPython wrapper (unbox and box happens here)
    - A Numba-generated cfunc wrapper to make the launcher callable from an njit/dpjit function

- [X] If this PR is a work in progress, are you filing the PR as a draft?


diptorupd:
TODOs

- [ ] Add a new debug option to generate printf calls from inside launcher function.
- [ ] Port the `NUMBA_DPEX_DUMP_KERNEL_LLVM` config option to experimental KernelDispatcher
- [ ] Add a new config option to dump the LLVM IR for the optimized launcher and cpython wrapper functions
- [ ] Generate proper mangled name for Range and NdRange arguments
- [ ] If the ndim attribute in the Range and NdRange data model not getting used during code generation, it should be removed

numba_dpex/experimental/launcher.py
```diff
@@ -0,0 +1,295 @@
...
+            num_flattened_kernel_args += 1
+
+    return num_flattened_kernel_args
+
+
+def _create_kernel_launcher_body(
+    codegen_targetctx: cpu.CPUContext,
+    kernel_targetctx: DpexKernelTargetContext,
+    builder: llvmir.IRBuilder,
+    indexer_argty: RangeType | NdRangeType,
```
ZzEeKkAa, line 89:
```suggestion
    indexer_argty: Union[RangeType, NdRangeType],
```
`Union` is available from `typing`:
```
from typing import Union
``
https://peps.python.org/pep-0604/

File: numba_dpex/experimental/launcher.py [hunk 2]
```diff
@@ -0,0 +1,297 @@
...
+        args_ty_list=args_ty_list,
+        datamodel_mgr=kernel_targetctx.data_model_manager,
+    )
+
+    if config.DEBUG_KERNEL_LAUNCHER:
+        cgutils.printf(
+            builder, "DPEX-DEBUG: Populated kernel args and arg type arrays.\n"
+        )
+
+    qref = _get_queue_ref_val(
```
ZzEeKkAa, line 138: What happens if we don't pass any dpnp arrays and allocates them inside dpjit?

numba_dpex/experimental/kernel_dispatcher.py
```diff
@@ -0,0 +1,324 @@
...
+                            )[1]
+
+                        raise e.bind_fold_arguments(folded)
+                    self.add_overload(kcres.cres_or_error, kcres.kernel_module)
+
+                # FIXME: enable caching
+
+                return kcres.kernel_module
+
+    def __getitem__(self, args):
```
ZzEeKkAa, line 285: Description mismatch. Accepts only one argument, and does not clones the objects. Also names are different.

numba_dpex/experimental/kernel_dispatcher.py
```diff
@@ -0,0 +1,324 @@
...
+from numba_dpex.core.descriptor import dpex_kernel_target
+from numba_dpex.core.exceptions import (
+    InvalidKernelLaunchArgsError,
+    UnsupportedKernelArgumentError,
+)
+from numba_dpex.core.kernel_interface.indexers import NdRange, Range
+from numba_dpex.core.pipelines import kernel_compiler
+from numba_dpex.core.types import DpnpNdArray
+
+_KernelLauncherLowerResult = namedtuple(
```
ZzEeKkAa, line 27: Dead code?

ZzEeKkAa (changes requested):
Changes look awesome!
It takes some time for me to understand what's going, so my review is still in progress.

numba_dpex/experimental/launcher.py
diptorupd: That is a good point. Somewhere we need to check if the array args are `DpnpNdArray`. Existing `JitKernel` has a check that I think I have not added anywhere.

numba_dpex/experimental/kernel_dispatcher.py
diptorupd: fixed.

numba_dpex/experimental/kernel_dispatcher.py
diptorupd: We do not need `__getitem__` and now it raised NotImplemented.

numba_dpex/experimental/launcher.py
diptorupd: I have added a check to evaluate queue equivalence before we compile.

diptorupd:
Merging the PR as it is an experimental preview feature. I have opened issues (refer: https://github.com/IntelPython/numba-dpex/issues/1120) to track all pending work and further polishing that needs to be done to make it ready for moving to `core`.

PR #1188, opened 2023-10-24
Fix lifetime management for sycl event

ZzEeKkAa:
Memory lifetime wasn't set properly which resulted in core dumps of running kernels if boxed sycl event argument where destroyed by GC. New implementation wraps parent into `MemInfo` object and lifetime of the object is managed by it which is properly managed by numba's compiler.

P.S. : also fixed functions signature for dpctl sycl binding.

### DIFF
--- a/numba_dpex/core/runtime/_dpexrt_python.c
+++ b/numba_dpex/core/runtime/_dpexrt_python.c
@@ -24,6 +24,8 @@
 #include "_queuestruct.h"
 #include "_usmarraystruct.h"
 
+#include "numba/core/runtime/nrt_external.h"
+
 // forward declarations
 static struct PyUSMArrayObject *PyUSMNdArray_ARRAYOBJ(PyObject *obj);
 static npy_intp product_of_shape(npy_intp *shape, npy_intp ndim);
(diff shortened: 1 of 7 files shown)


numba_dpex/core/runtime/_dpexrt_python.c
```diff
@@ -1328,7 +1334,9 @@ static int DPEXRT_sycl_event_from_python(PyObject *obj,
         goto error;
     }
 
-    event_struct->parent = obj;
+    Py_INCREF(event_obj);
```
diptorupd, line 1337: Please add a comment here. The `Py_INCREF` is to ensure that Numba lifetime analysis and cleanups do not impact the original Python object that was passed into Numba.

numba_dpex/dpctl_iface/libsyclinterface_bindings.py
```diff
@@ -71,7 +71,7 @@ def dpctl_event_wait(builder: llvmir.IRBuilder, *args):
     mod = builder.module
     fn = _build_dpctl_function(
         llvm_module=mod,
-        return_ty=cgutils.voidptr_t,
+        return_ty=llvmir.VoidType(),
```
diptorupd, line 74: These are unrelated changes. I have also made the same changes in https://github.com/IntelPython/numba-dpex/pull/1178. 

The changes should be in a separate commit no need to squash into the SyclEvent changes commit

diptorupd (approved):
LGTM! Thank you for these important changes.

PR #1190, opened 2023-10-25
Fix lifetime management for sycl queue

ZzEeKkAa:
Memory lifetime wasn't set properly which resulted in core dumps of running kernels if boxed sycl event argument where destroyed by GC. New implementation wraps parent into `MemInfo` object and lifetime of the object is managed by it which is properly managed by numba's compiler.

Similar changes to https://github.com/IntelPython/numba-dpex/pull/1188

Closes https://github.com/IntelPython/numba-dpex/issues/1180


diptorupd (approved):
Thank you very much!

PR #1194, opened 2023-11-02
Numba-mlir integration

Hardcode84:
This is an alternative to https://github.com/IntelPython/numba-dpex/pull/1130

Instead of using target options, use an alternative pipeline class.
Options code is still there so we can switch after they fixed.


ZzEeKkAa:
I've converted PR to draft, since it is WIP. @Hardcode84 are you going to update CI in this PR? 
Some highlights from our call:
- Add `NUMBA_DPEX_DEFAULT_JIT_COMPILER` that can be `mlir`
- Update testing matrix to run tests with and without mlir

diptorupd:
Overall the changes look good and are compartmentalized we should merge once we have a proper plan for CI.

ZzEeKkAa (approved):
LGTM! Thank you!

Issue #1106, opened 2023-08-02
This kernel outputs wrong results on CPU, featuring nested kernel funcs, private and local memories, and barriers


fcharras:
I recently bumped all dependencies from our KMeans project with numba_dpex at https://github.com/soda-inria/sklearn-numba-dpex/ , including bump to OneAPI 2023.2.0 , latest numba_dpex, numba>=0.57, drivers, etc.

A few tests that used to be stable before started to fail when ran on CPU, the cause is that some kernels output wrong values. It has similar expression and consequences than https://github.com/IntelPython/numba-dpex/issues/906 . Those issues really shake the core of what developers need to embrace the stack and build more complex system on top. I hope the reproducers I try to craft could unlock progress on this side.

The reproducer is a fairly complicated kernel with nested kernel funcs, using private and local memories, and barriers. I can't seem to manage to reduce it further to a simpler reproducer, so there it is:

<details>
  <summary>Click to expand the reproducer</summary>

```python
import numba_dpex as dpex
import dpctl.tensor as dpt

dtype = dpt.float32

nb_Y_window_cols = 8
nb_Y_window_rows = 10
Y_window_shape = nb_Y_window_rows, nb_Y_window_cols


@dpex.kernel
def columnwise_broadcast_dot_product(
    X,             # IN, shape (10)
    Y,             # IN, shape (10, 10)
    dot_products,  # IN  shape (10)
):
    """Write into dot_products the results of each dot product of X with a 
    column of Y
    """

    # All work items will cooperate to load windows of Y data into shared
    # memory. Each work item is responsible for loading a single slot in
    # shared memory, at following coordinates:
    Y_local_row_idx = dpex.get_local_id(1)
    Y_local_col_idx = dpex.get_local_id(0)


    # The only work item that will actually compute the dot products and write
    # them to global memory is the first fork item
    is_first_work_item = (dpex.get_global_id(0) == 0) and (dpex.get_global_id(1) == 0)

    # allocate shared memory where windows of Y will be cooperateively loaded
    Y_window = dpex.local.array(shape=Y_window_shape, dtype=dtype)

    # allocate private memory where dot products will be accumulated by the
    # first work item
    partial_dot_products = dpex.private.array(shape=8, dtype=dtype)

    first_Y_col_idx = 0

    for Y_window_idx in range(2):  # Y has 10 cols so is covered by 2 * 8 cols
        # NB: to ensure no issue related to
[...]

PR #1204, opened 2023-11-04
Fix barrier issue

Hardcode84:
LLVM optimizer was tranforming code
```
i = get_global_id(0)
if i < 1:
    arr1[0] = ...

barrier()

if i < 0:
    arr2[0] = ...
```
into
```
i = get_global_id(0)
if i < 1:
    arr1[0] = ...
    barrier()
    arr2[0] = ...
else:
    barrier()
```
Which is valid transform for LLVM IR, but is UB per OpenCL spec.
Add `noduplicate` attr to barrier funcs to avoid this transform.

### DIFF
--- a/numba_dpex/ocl/oclimpl.py
+++ b/numba_dpex/ocl/oclimpl.py
@@ -105,6 +105,7 @@ def barrier_one_arg_impl(context, builder, sig, args):
     barrier = _declare_function(
         context, builder, "barrier", sig, ["unsigned int"]
     )
+    barrier.attributes.add("noduplicate")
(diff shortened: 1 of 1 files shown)


diptorupd:
@Hardcode84 thank you for giving the main intuition for fixing the barrier issue. After further analysis (refer #1106) we are now adding the `convergent` attribute to prevent the transformation you highlighted. In addition, we added a new `group_barrier` call that works only for `NdRange` kernels (only in `numba_dpex.experimental` for now). Together, these two have fixed the code generation problems with barriers in numba_dpex. 

I am closing your PR.

PR #1210, opened 2023-11-07
Switch to sycl compiler

ZzEeKkAa:


chudur-budur (approved):
LGTM!

Note: The two jenkins tests that are failing are from `dpctl`, @oleksandr-pavlyk has already upstreamed the fix, if I am correct.

conda-recipe/bld.bat
```diff
@@ -3,6 +3,20 @@
 @REM used BUILD_PREFIX as compiler installed in build section of meta.yml
 set "PATH=%BUILD_PREFIX%\Library\bin-llvm;%PATH%"
 
+REM A workaround for activate-dpcpp.bat issue to be addressed in 2021.4
```
diptorupd, line 6: We are almost in 2024.1 is the workaround still applicable?

ZzEeKkAa: It is there for dpctl, dpnp and dpbench. Even if it is not the case anymore, I would prefer to be consistent

ZzEeKkAa: And since CI debugging takes lot of time, I would prefer not to experiment with it.

chudur-budur:
If every issue is resolved, can we merge it? @ZzEeKkAa @diptorupd

PR #1212, opened 2023-11-09
Move versioneer and pytest configs into pyproject.toml

diptorupd:
We were using `setup.cfg` to define the configuration options for `versioneer` and `pytest`. Both of these can be moved into `pyproject.toml`. 

The PR moves the configurations from `setup.cfg` to `pyproject.toml` and removes `setup.cfg`.


ZzEeKkAa:
`Generate coverage and push to Coveralls.io` is not triggered because of:
```yaml
name: Coverage
on:
  push:
    branches:
      - main
  pull_request:
    paths:
      - 'numba_dpex/**'
      - environment/coverage.yml
 ```
 
It simply is not triggered...

**Comment by diptorupd** [comment]
> UPD: why pytest kept in `setup.cfg` ?

I was experimenting, to see what changed to impact the coverage workflow.

File: pyproject.toml [hunk 1]
```diff
@@ -60,3 +61,24 @@ requires = [
     "versioneer-518"
 ]
 build-backend = "setuptools.build_meta"
+
+[tool.pytest.ini.options]
```
ZzEeKkAa, line 65:
```suggestion
[tool.pytest.ini_options]
```
https://docs.pytest.org/en/stable/reference/customize.html#pyproject-toml

pyproject.toml
```diff
@@ -60,3 +61,24 @@ requires = [
...
+
+[tool.pytest.ini_options]
+minversion = "6.0"
+norecursedirs= [
+    ".*", "*.egg*", "build", "dist", "conda-recipe",
+]
+xfail_strict = true
+addopts = [
+    "--disable-warnings",
+    "--ignore setup.py",
```
ZzEeKkAa, line 73: Don't you need to make them separate strings?
```suggestion
    "--ignore", "setup.py",
```

Same for others.

ZzEeKkAa: or may be just for the last line, I'm not sure, but the workflow is failing

pyproject.toml
```diff
@@ -60,3 +61,19 @@ requires = [
...
+minversion = "6.0"
+norecursedirs= [
+    ".*", "*.egg*", "build", "dist", "conda-recipe",
+]
+xfail_strict = true
+addopts = [
+    "--disable-warnings",
+    "--strict",
+    "--durations=20",
+    "-ra -vv",
```
ZzEeKkAa, line 75: This option set through the command line, same for disable-warnings

diptorupd: Removed the arguments that are specified in `pyproject.toml` from the CLI specified in the workflow when invoking `pytest`.

ZzEeKkAa (approved):
LGTM! Thank you!

PR #1213, opened 2023-11-11
Add a target inside the experimental module

diptorupd:
Changes introduced in the experimental module should not have side effects on the `numba_dpex.core` components. For this reason, an experimental target is being introduced into `numba_dpex.experimental`. Any future typing or lowering changes for any
experimental feature should be introduced here.

### DIFF
--- a/numba_dpex/experimental/kernel_dispatcher.py
+++ b/numba_dpex/experimental/kernel_dispatcher.py
@@ -13,6 +13,7 @@ from numba.core import errors, sigutils, types
 from numba.core.compiler import CompileResult
 from numba.core.compiler_lock import global_compiler_lock
 from numba.core.dispatcher import Dispatcher, _FunctionCompiler
+from numba.core.target_extension import dispatcher_registry, target_registry
 from numba.core.typing.typeof import Purpose, typeof
 
 from numba_dpex import config, spirv_generator
@@ -315,3 +316,7 @@ class KernelDispatcher(Dispatcher):
         """Functor to launch a kernel."""
 
         raise NotImplementedError
+
+
+_dpex_target = target_registry["dpex_kernel"]
+dispatcher_registry[_dpex_target] = KernelDispatcher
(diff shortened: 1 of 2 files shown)


numba_dpex/experimental/kernel_dispatcher.py
```diff
@@ -13,17 +13,19 @@
...
 from numba_dpex import config, spirv_generator
-from numba_dpex.core.descriptor import dpex_kernel_target
 from numba_dpex.core.exceptions import (
     ExecutionQueueInferenceError,
     UnsupportedKernelArgumentError,
 )
 from numba_dpex.core.pipelines import kernel_compiler
 from numba_dpex.core.types import DpnpNdArray
 
+from .target import dpex_exp_kernel_target
```
ZzEeKkAa, line 27: do we want to avoid relative imports?

numba_dpex/experimental/kernel_dispatcher.py
```diff
@@ -177,7 +179,7 @@ class KernelDispatcher(Dispatcher):
 
     """
 
-    targetdescr = dpex_kernel_target
+    targetdescr = dpex_exp_kernel_target
```
ZzEeKkAa, line 182: I can't see usage of `targetdescr`

diptorupd: It is needed by the superclass: https://github.com/numba/numba/blob/a4664180ddc91e4c8a37dd06324f4e229a76df91/numba/core/dispatcher.py#L796

The variable is used in the `super().__init__`: https://github.com/numba/numba/blob/a4664180ddc91e4c8a37dd06324f4e229a76df91/numba/core/dispatcher.py#L823

numba_dpex/experimental/kernel_dispatcher.py
diptorupd: We can, but in the past I have found relative imports to be the only way out of avoiding circular imports.

ZzEeKkAa: It is okay for now then. We just need to rethink design to avoid circular imports

ZzEeKkAa (approved):
LGTM!

numba_dpex/experimental/kernel_dispatcher.py
diptorupd: I agree 👍

PR #1227, opened 2023-11-22
Adds a new literal type to store IntEnum as Literal types.

diptorupd:
    - Adds a new `IntEnumLiteral` type with corresponding data model into the `DpexExpKernelTargetContext`.
      The type is used to pass in or define an `IntEnum` flag as an `numba.types.IntegerLiteral` inside a kernel function.

### DIFF
--- a/numba_dpex/experimental/models.py
+++ b/numba_dpex/experimental/models.py
@@ -17,7 +17,7 @@ from .literal_intenum_type import IntEnumLiteral
 from .types import KernelDispatcherType
 
 
-class LiteralIntEnumModel(PrimitiveModel):
+class IntEnumLiteralModel(PrimitiveModel):
     """Representation of an object of LiteralIntEnum type using Numba's
     PrimitiveModel that can be represented natively in the target in all
     usage contexts.
@@ -42,7 +42,7 @@ def _init_exp_data_model_manager() -> DataModelManager:
     dmm = dpex_core_models.dpex_data_model_manager.copy()
 
     # Register the types and data model in the DpexExpTargetContext
-    dmm.register(IntEnumLiteral, LiteralIntEnumModel)
+    dmm.register(IntEnumLiteral, IntEnumLiteralModel)
 
     return dmm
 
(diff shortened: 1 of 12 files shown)


diptorupd:
@ZzEeKkAa the PR is good to go as-is. There is an additional unit test case I want to add that will verify that an `intEnumLiteral` type is parsed as an constant literal and lowered as such in the LLVM IR. I will add that test case after I have fixed the overload inlining in a separate PR.

numba_dpex/experimental/models.py
```diff
@@ -6,14 +6,28 @@
...
+from numba.core.datamodel.models import PrimitiveModel
 from numba.core.extending import register_model
 
 import numba_dpex.core.datamodel.models as dpex_core_models
 
+from .literal_intenum_type import IntEnumLiteral
 from .types import KernelDispatcherType
 
 
+class LiteralIntEnumModel(PrimitiveModel):
```
ZzEeKkAa, line 20: Should we be consistent with names? `<class>` and `<class>Model` sounds better for me.

ZzEeKkAa:
> @ZzEeKkAa the PR is good to go as-is. There is an additional unit test case I want to add that will verify that an `intEnumLiteral` type is parsed as an constant literal and lowered as such in the LLVM IR. I will add that test case after I have fixed the overload inlining in a separate PR.

May be we can add this test case here and mark it as xfail? Don't want you to overload inline fix with unrelated code.

numba_dpex/experimental/models.py
diptorupd: Done

diptorupd:
> May be we can add this test case here and mark it as xfail? Don't want you to overload inline fix with unrelated code.

The test case I wanted to add was to make sure that `FlagEnum` objects are lowered as scalar values in the generated LLVM IR. I have added the test in [ec3e962](https://github.com/IntelPython/numba-dpex/pull/1227/commits/ec3e962a25e6b1308cb672dcfedabf9fd9ff3042). The PR is good to go and I will merge after CI is happy.

PR #1230, opened 2023-11-25
Enables adding overload to DpexExpKernelTarget and fully inline them into the final module.

diptorupd:
Add a new `device_func` decorator to compile `DpexKernelTarget` overloads 
   - A new `generate_device_ir` option for `DpexTargetDescriptor` is introduced. The new option can be used to prevent a module to be compiled to device IR binary.
    - A new `device_func` decorator that is registered to compile overloads in `DpexExpKernelTarget`. The `device_func` decorated
      functions are not compiled to device IR.
    - The `kernel` decorator is updated to compile a finalized module to device IR.

### DIFF
--- a/numba_dpex/core/descriptor.py
+++ b/numba_dpex/core/descriptor.py
@@ -41,6 +41,7 @@ class DpexTargetOptions(CPUTargetOptions):
     release_gil = _option_mapping("release_gil")
     no_compile = _option_mapping("no_compile")
     use_mlir = _option_mapping("use_mlir")
+    inline_threshold = _option_mapping("inline_threshold")
     _compilation_mode = _option_mapping("_compilation_mode")
 
     def finalize(self, flags, options):
@@ -49,6 +50,7 @@ class DpexTargetOptions(CPUTargetOptions):
         _inherit_if_not_set(flags, options, "release_gil", False)
         _inherit_if_not_set(flags, options, "no_compile", True)
         _inherit_if_not_set(flags, options, "use_mlir", False)
+        _inherit_if_not_set(flags, options, "inline_threshold", 0)
         _inherit_if_not_set(
             flags, options, "_compilation_mode", CompilationMode.KERNEL
         )
--- a/numba_dpex/core/descriptor.py
+++ b/numba_dpex/core/descriptor.py
@@ -11,6 +11,7 @@ from numba.core.descriptors import TargetDescriptor
 from .targets.dpjit_target import DPEX_TARGET_NAME, DpexTargetContext
 from .targets.kernel_target import (
     DPEX_KERNEL_TARGET_NAME,
+    CompilationMode,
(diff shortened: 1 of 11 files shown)


diptorupd:
Unit tests design sketch

- [ ] Basic overload end to end
- [ ] compilation_mode option in decorator (negative test case)
- [ ] check if overloads are available in DpexExpKerneltarget._defns and no other target
- [ ] check if overload only compiled to LLVM bitcode
- [ ] check if overload call generated in kernel
- [ ] kernel return error (copy from existing core test case)

- [ ] inlining : check if can be set using decorator
- [ ] inlining : check if overload gets fully inlined or not (two cases) (LLVM IR level)

numba_dpex/core/descriptor.py
```diff
@@ -40,13 +41,19 @@ class DpexTargetOptions(CPUTargetOptions):
...
+    inline_threshold = _option_mapping("inline_threshold")
+    _compilation_mode = _option_mapping("_compilation_mode")
 
     def finalize(self, flags, options):
         super().finalize(flags, options)
         _inherit_if_not_set(flags, options, "experimental", False)
         _inherit_if_not_set(flags, options, "release_gil", False)
         _inherit_if_not_set(flags, options, "no_compile", True)
         _inherit_if_not_set(flags, options, "use_mlir", False)
+        _inherit_if_not_set(flags, options, "inline_threshold", 0)
```
ZzEeKkAa, line 53:
```suggestion
        _inherit_if_not_set(flags, options, "inline_threshold", config.INLINE_THRESHOLD)
```

Let's set it here based on the config value to avoid code duplicate and complex logic in `_optimize_final_module`. Should we warn if both options set, so user knows that env variable does not take effect on particular function?

numba_dpex/experimental/kernel_dispatcher.py
```diff
@@ -81,10 +84,18 @@ def _compile_to_spirv(
...
             kernel_func, kernel_fndesc.argtypes
         )
-
-        # makes sure that the spir_func is completely inlined into the
-        # spir_kernel wrapper
-        kernel_library.optimize_final_module()
+        # If the inline_threshold option was set then set the property in the
+        # kernel_library to force inlining ``overload`` calls into a kernel.
+        try:
+            kernel_library.inline_threshold = self.targetoptions[
```
ZzEeKkAa, line 90:
```suggestion
            kernel_library.inline_threshold = self.targetoptions.get(
```
Should we use `get` instead of try-catch?

diptorupd: Yes, I was not aware of it. Will make the change.

diptorupd: Addressed

adarshyoga (approved):

PR #1231, opened 2023-11-30
Clean up experimental launcher

ZzEeKkAa:
Code restructuring to achieve encapsulation and simplify logic. Some comments where added for easier navigation.

It is not perfect, but better than it was

### DIFF
--- a/numba_dpex/experimental/launcher.py
+++ b/numba_dpex/experimental/launcher.py
@@ -6,19 +6,19 @@
 from either CPython or a numba_dpex.dpjit decorated function.
 """
 
-from collections import namedtuple
-from typing import Union
+from typing import NamedTuple, Union
 
 import dpctl
 from llvmlite import ir as llvmir
 from numba.core import cgutils, cpu, types
 from numba.core.datamodel import default_manager as numba_default_dmm
+from numba.core.types.containers import UniTuple
+from numba.core.types.functions import Dispatcher
 from numba.extending import intrinsic
 
 from numba_dpex import config, dpjit, utils
 from numba_dpex.core.exceptions import UnreachableError
 from numba_dpex.core.runtime.context import DpexRTContext
-from numba_dpex.core.targets.kernel_target import DpexKernelTargetContext
 from numba_dpex.core.types import (
     DpctlSyclEvent,
     DpnpNdArray,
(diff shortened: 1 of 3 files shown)


numba_dpex/experimental/launcher.py
```diff
@@ -28,153 +27,49 @@
...
         builder: llvmir.IRBuilder,
     ):
-        self._cpu_codegen_targetctx = codegen_targetctx
-        self._kernel_targetctx = kernel_targetctx
-        self._builder = builder
-        if kernel_targetctx:
-            self._klbuilder = kl.KernelLaunchIRBuilder(
-                kernel_targetctx, builder
-            )
+        # No computations here. ALl IR code that is generated by any builder
```
diptorupd, line 47: It is not clear to me what this comment means.

numba_dpex/experimental/launcher.py
```diff
@@ -28,153 +27,49 @@
...
-            callargs_ptrs=kernel_args_ptrs,
-            kernel_argtys=kernel_argtys,
-            args_list=args_list,
-            args_ty_list=args_ty_list,
-            datamodel_mgr=self._kernel_targetctx.data_model_manager,
-        )
+        ty_kernel_args_tuple,
+        ll_kernel_args_tuple,
+    ):
+        """Convert arguments to kernel arguments because kernel and dpjit use
```
diptorupd, line 63: Same here. The docstring does not seem to match what the function is doing.

numba_dpex/experimental/launcher.py
```diff
@@ -229,70 +125,85 @@ def get_queue_ref_val(
...
+        # Inserts a global constant byte string in the current LLVM module to
+        # store the passed in SPIR-V binary blob.
+        kernel_bc_byte_str = self.context.insert_const_bytes(
+            self.builder.module,
+            bytes=kernel_module.kernel_bitcode,
         )
-        return sycl.dpctl_kernel_bundle_get_kernel(
-            self._builder, kbref, kernel_name
+
+        # Send it to dpctl, so it create this function on device and returns
```
diptorupd, line 148: How about:

Create a sycl::kernel_bundle object and return it as an opaque pointer using dpctl's libsyclinterface.

numba_dpex/experimental/launcher.py
```diff
@@ -229,70 +125,85 @@ def get_queue_ref_val(
...
     def create_llvm_values_for_index_space(
         self,
-        indexer_argty: Union[RangeType, NdRangeType],
-        index_space_arg: llvmir.BaseStructType,
-    ) -> _LLVMIRValuesForIndexSpace:
-        """Returns a list of LLVM IR Values that hold the unboxed extents of a
-        Python Range or NdRange object.
+        ty_indexer_arg: Union[RangeType, NdRangeType],
+        index_arg: llvmir.BaseStructType,
+    ) -> tuple[list, list]:
```
diptorupd, line 172: Why are you changing from a namedtuple to tuple? Returning a namedtuple makes the code much more clearer to follow as we are basically returning a instance of a class that will give a clear idea of the fields and allow accessing the members in a clean way.

numba_dpex/experimental/launcher.py
```diff
@@ -416,74 +289,71 @@ def _submit_kernel(
...
+            global_range_extents,
+            local_range_extents,
+        ) = generator.create_llvm_values_for_index_space(
+            ty_indexer_arg=ty_index_space,
+            index_arg=ll_index_space,
         )
 
-        kref = fn_body_gen.get_kernel(kernel_module, kbref)
-
-        index_space_values = fn_body_gen.create_llvm_values_for_index_space(
```
diptorupd, line 467: If you are changing namedtuple to a tuple to get the IDE to better statically analyze the code, better add a type annotation here.

numba_dpex/experimental/launcher.py
```diff
@@ -28,153 +27,49 @@
...
-    "_LLVMIRValuesForNdRange", ["global_range_extents", "local_range_extents"]
+from numba_dpex.experimental.kernel_dispatcher import (
+    KernelDispatcher,
+    _KernelModule,
 )
+from numba_dpex.utils import create_null_ptr
 
 
-class _LaunchTrampolineFunctionBodyGenerator:
+class _LaunchKernelIRGenerator:
```
diptorupd, line 37: The class name is not clear. The previous name was specific to the class being the helper to build the `launch_trampoline` intrinsic function's body.

How about: _SubmitKernelCo

numba_dpex/experimental/launcher.py
```diff
@@ -28,153 +27,49 @@
...
     """
     Helper class to generate the LLVM IR for the launch_trampoline intrinsic.
```
diptorupd, line 39: The doc string no longer applies since the launch_trampoline intrinsic is gone.

numba_dpex/experimental/launcher.py
```diff
@@ -303,110 +214,72 @@ def create_kernel_bundle_from_spirv(
...
-        sycl.dpctl_kernel_bundle_delete(self._builder, kernel_bundle_ref)
-
 
 @intrinsic(target="cpu")
 def _submit_kernel(
     typingctx,  # pylint: disable=W0613
-    kernel_fn,
-    index_space,
-    kernel_args,
+    ty_kernel_fn: Dispatcher,
```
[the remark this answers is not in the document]
diptorupd: hmm `KernelDispatcherType` is a subclass of `Dispatcher`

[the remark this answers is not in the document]
ZzEeKkAa: I've just tried it with `@overload`, but still no difference...

numba_dpex/experimental/launcher.py
ZzEeKkAa: I'm actually thinking of getting rid of this class and use functions instead. What do you think? I'm thinking about it, since class is used as builder and then get's destroyed within one function body.

numba_dpex/experimental/launcher.py
ZzEeKkAa: My biggest concern was that it does not provide datatype. I've figured out that it is possible to define it in different way:

```python
class class_name(NamedTuple):
    field1: datatype
    field2: datatype
```

That is equivalent to 
```python
class_name = collections.namedtuple('class_name', ['field1', 'field2'])
```
Source: https://www.geeksforgeeks.org/typing-namedtuple-improved-namedtuples/

ZzEeKkAa: I have nothing against using named tuple if they make sense and does not affect readability/static analysis. In this case it makes sense, but passing named tuple instead of list of arguments seems to be redundant...

numba_dpex/experimental/launcher.py
```diff
@@ -416,74 +289,71 @@ def _submit_kernel(
...
-            kernel_args=kernel_args_unpacked,
-        )
+        # creates new object, so we must clean it up
+        kernel_ref = generator.get_kernel(queue_ref, kernel_module)
 
-        kbref = fn_body_gen.create_kernel_bundle_from_spirv(
-            queue_ref=qref,
-            kernel_bc=kernel_bc_byte_str,
-            kernel_bc_size_in_bytes=len(kernel_module.kernel_bitcode),
+        # What is it for?
```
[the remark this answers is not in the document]
ZzEeKkAa: Yeap. Actually was thinking if we need to do it here or move it inside `KernelLaunchIRBuilder`

numba_dpex/experimental/launcher.py
diptorupd: My biggest pet peeve is that not every code where a function returns a tuple is written in a disciplined way like you originally had.

I have seen way too many places where someone did `result = function_returns_tuple(); a = result[0]; b = result[1]` making the code almost unreadable, especially if the `function_returns_tuple` is deep inside a library.

numba_dpex/experimental/launcher.py
diptorupd: No strong opinion, tbh. We will need to pass two extra args to each function and removing the class kind of looses the logical relation between the functions that they all are doing something related. But, if you want to change it go ahead.

diptorupd (approved):
Thank you! The changes made the launcher module better organized than before. We are laying a good foundation for these new features.

Issue #1070, opened 2023-06-14
Overhead on running dpex kernels


AlexanderKalistratov:
Reproducer:
```python
import dpnp
import time
import numba_dpex
from numba_dpex import Range,NdRange

@numba_dpex.kernel
def bar(a):
    a[0] = 0

a = dpnp.empty(1)
for i in range(10):
    start = time.time()
    bar[Range(1)](a)
    end = time.time()
    print(1000*(end - start))
```

Result:
```
170.96805572509766
0.9398460388183594
0.7669925689697266
0.7381439208984375
0.6814002990722656
0.7398128509521484
0.5812644958496094
0.6299018859863281
0.7305145263671875
0.7123947143554688
```

Expected result... about 0.1ms.

More details here: https://github.com/IntelPython/numba-dpex/pull/946#issuecomment-1513849247

diptorupd:
Addressed in #1236.

The new `numba_dpex.experimental.kernel` uses a fully native kernel dispatch mechanism that fixes the previous overhead reported in this issue.

On my local test machine, existing `numba_dpex.kernel` reported following performance numbers:

```bash
518.5830593109131
5.5999755859375
5.43522834777832
5.475282669067383
5.715131759643555
5.93256950378418
5.736351013183594
5.846977233886719
5.640983581542969
6.529092788696289
```

Using the `numba_dpex.experimental.kernel` from PR #1236. The following timing was captured:

```bash
236.9523048400879
0.2722740173339844
0.14162063598632812
0.16736984252929688
0.16379356384277344
0.16546249389648438
0.18072128295898438
0.13518333435058594
0.13208389282226562
0.15735626220703125
```

PR #1236, opened 2023-12-07
Enable SPIR-V to device caching

ZzEeKkAa:
Add kernel reference caching for the SPIRV to device compilation. This is a critical change to make experimental kernels usable.

Should we remove comment from `kernel_dispatcher.py`?
```python
# TODO: Enable caching
# Add code to enable on disk caching of a binary spirv kernel.
# Refer: https://github.com/IntelPython/numba-dpex/issues/1197
```
I've add simple test to check if second run does not crash XD. I've made sure that kernel reference is taking from cache by enabling logs and making sure it reached right place.

Since we can not implement caching on kernel dispatcher level due to missing information about context that is available only at runtime, I guess we can close this issue.

Closes: #1070


diptorupd:
> Should we remove comment from `kernel_dispatcher.py`?
```python
# TODO: Enable caching
# Add code to enable on disk caching of a binary spirv kernel.
# Refer: https://github.com/IntelPython/numba-dpex/issues/1197
```

No, that is a different issue. Numba supports ahead of time (AOT) compilation using on disk caching of compiled binaries. That way if you run a function with `cache=True` flag the function gets serialized to disk and the next time Numba is started the on disk cached function is de-serialized and used. In our case, that will mean we serialize the `SPIRV` binary generated for a kernel to disk and next time de-serialize and use it. KernelBundle caching comes after that and unrelated.

diptorupd (approved):
Thank you! One step closer to making the experimental kernel really usable. :)

PR #1238, opened 2023-12-11
Removes code duplication and adds new unit tests.

diptorupd:
    - Removes duplicate code inside codegen module.
    - Adds a unit test to make sure a compiler warning is generated
      when the `inline_threshold` is set to >=3

### DIFF
--- a/numba_dpex/core/codegen.py
+++ b/numba_dpex/core/codegen.py
@@ -70,7 +70,13 @@ class SPIRVCodeLibrary(CPUCodeLibrary):
             )
 
         pmb.disable_unit_at_a_time = False
-        pmb.inlining_threshold = self.inline_threshold
+
+        # The PassManagerBuilder's inlining_threshold property is set only when
+        # inline_threshold is g.t. 0. Doing otherwise, *i.e.*, setting the
+        # pmb.inlining_threshold to 0 will lead to at minimum `alwaysinline`
+        # pass to run.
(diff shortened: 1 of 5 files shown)


numba_dpex/core/codegen.py
```diff
@@ -55,44 +57,22 @@ def _optimize_final_module(self):
...
-        else:
-            if (
-                hasattr(self, "_inline_threshold")
-                and self._inline_threshold > 0
-                and self._inline_threshold <= 3
-            ):
-                logging.warning(
-                    "Setting INLINE_THRESHOLD leads to very aggressive "
-                    + "optimizations that may produce incorrect binary."
+        if hasattr(self, "_inline_threshold") and self._inline_threshold > 0:
```
ZzEeKkAa, line 68:
```suggestion
        pmb.inlining_threshold = self.inline_threshold
```
why not to set it unconditionally? We can warn in the `@inline_threshold.setter`

ZzEeKkAa (approved):
LGMT! Thank you. Just one nitpicking

numba_dpex/core/codegen.py
diptorupd: Makes sense. Done

ZzEeKkAa (approved):
LGTM! Thank you!

ZzEeKkAa (approved):

PR #1243, opened 2023-12-13
Apply builder pattern to kernel launcher

ZzEeKkAa:
Update kernel launcher builder to use builder pattern. Since the argument list is long we need modular way of providing those arguments to make code more readable and extendable.

### DIFF
--- a/numba_dpex/core/parfors/parfor_lowerer.py
+++ b/numba_dpex/core/parfors/parfor_lowerer.py
@@ -13,12 +13,15 @@ from numba.parfors.parfor import (
 )
 
 from numba_dpex import config
-from numba_dpex.core.datamodel.models import dpex_data_model_manager as dpex_dmm
 from numba_dpex.core.parfors.reduction_helper import (
     ReductionHelper,
     ReductionKernelVariables,
 )
 from numba_dpex.core.utils.kernel_launcher import KernelLaunchIRBuilder
+from numba_dpex.dpctl_iface import libsyclinterface_bindings as sycl
+from numba_dpex.core.datamodel.models import (
+    dpex_data_model_manager as kernel_dmm,
+)
 
 from ..exceptions import UnsupportedParforError
 from ..types.dpnp_ndarray_type import DpnpNdArray
@@ -26,11 +29,6 @@ from .kernel_builder import ParforKernel, create_kernel_for_parfor
 from .reduction_kernel_builder import (
     create_reduction_main_kernel_for_parfor,
     create_reduction_remainder_kernel_for_parfor,
-)
-
-_KernelArgs = namedtuple(
-    "_KernelArgs",
-    ["num_flattened_args", "arg_vals", "arg_types"],
 )
 
 
@@ -68,11 +66,8 @@ def _getvar(lowerer, x):
         var_val = lowerer.varmap[x]
 
     if var_val:
-        if not isinstance(var_val.type, llvmir.PointerType):
-            with lowerer.builder.goto_entry_block():
-                var_val_ptr = lowerer.builder.alloca(var_val.type)
-            lowerer.builder.store(var_val, var_val_ptr)
-            return var_val_ptr
+        if isinstance(var_val.type, llvmir.PointerType):
+            return lowerer.builder.load(var_val)
         else:
             return var_val
     else:
@@ -90,56 +85,6 @@ class ParforLowerImpl:
     """Provides a custom lowerer for parfor nodes that generates a SYCL kernel
     for a parfor and submits it to a queue.
     """
-
-    def _build_kernel_arglist(
-        self, kernel_fn, lowerer, kernel_builder: KernelLaunchIRBuilder
-    ):
-        """Creates local variables for all the arguments and the argument types
-        that are passes to the kernel function.
-
-        Args:
-            kernel_fn: Kernel function to be launched.
-            lowerer: The Numba lowerer used to generate the LLVM IR
-
-        Raises:
-            AssertionError: If the LLVM IR Value for an argument defined in
-            Numba IR is not found.
-        """
-        num_flattened_args = 0
-
-        # Compute number of args to be passed to the kernel. Note that the
-        # actual number of kernel arguments is greater than the count of
-        # kernel_fn.kernel_args as arrays get flattened.
-        for arg_type in kernel_fn.kernel_arg_types:
-            if isinstance(arg_type, DpnpNdArray):
-                datamodel = dpex_dmm.lookup(arg_type)
-                num_flattened_args += datamodel.flattened_field_count
-            elif arg_type == types.complex64 or arg_type == types.complex128:
-                num_flattened_args += 2
-            else:
-                num_flattened_args += 1
-
-        # Create LLVM values for the kernel args list and kernel arg types list
-        args_list = kernel_builder.allocate_kernel_arg_array(num_flattened_args)
-        args_ty_list = kernel_builder.allocate_kernel_arg_ty_array(
-            num_flattened_args
-        )
-        callargs_ptrs = []
-        for arg in kernel_fn.kernel_args:
-            callargs_ptrs.append(_getvar(lowerer, arg))
-
-        kernel_builder.populate_kernel_args_and_args_ty_arrays(
-            kernel_argtys=kernel_fn.kernel_arg_types,
-            callargs_ptrs=callargs_ptrs,
-            args_list=args_list,
-            args_ty_list=args_ty_list,
-        )
-
-        return _KernelArgs(
-            num_flattened_args=num_flattened_args,
-            arg_vals=args_list,
-            arg_types=args_ty_list,
-        )
 
     def _loop_ranges(
         self,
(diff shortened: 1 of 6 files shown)


numba_dpex/core/utils/kernel_launcher.py
```diff
@@ -68,15 +116,15 @@ def _build_array_attr_arg(
...
+        self._build_arg(
             val=array_attr,
             ty=array_attr_ty,
             arg_list=arg_list,
             args_ty_list=args_ty_list,
             arg_num=arg_num,
         )
 
-    def _build_unituple_member_arg(
+    def _build_unituple_member_arg(  # pylint: disable=too-many-arguments
```
diptorupd, line 127: Adding the pylint disabling is too much noise in code. Following linter guide line is OK, but there should be room to tweak.
For this particular check , can you update the `pylintrc` max-args to 6.

numba_dpex/core/parfors/parfor_lowerer.py
```diff
@@ -68,11 +69,8 @@ def _getvar(lowerer, x):
         var_val = lowerer.varmap[x]
 
     if var_val:
-        if not isinstance(var_val.type, llvmir.PointerType):
-            with lowerer.builder.goto_entry_block():
-                var_val_ptr = lowerer.builder.alloca(var_val.type)
-            lowerer.builder.store(var_val, var_val_ptr)
-            return var_val_ptr
+        if isinstance(var_val.type, llvmir.PointerType):
```
diptorupd, line 72: Why is this change needed? Is `var_val` always a `PointerType`?

numba_dpex/core/utils/kernel_launcher.py
ZzEeKkAa: I don't want to apply global settings for the number of max-args cause it affects readability. It is a todo work to update this code to use less arguments.

numba_dpex/core/parfors/parfor_lowerer.py
ZzEeKkAa: It used to return pointer to value, not it returns value as description says.

ZzEeKkAa: Change needed in parfor to be aligned with the semantic of numba's unboxed llvm values. Resulted llvmir is not affected, since value to ptr concertation is done in another step ( `KernelLauncerIRBuilder.set_arguments` ).

ZzEeKkAa: It also makes sense because `ParforKernel.kernel_args_ty` now actually match types of the llvm values.

numba_dpex/core/parfors/parfor_lowerer.py
```diff
@@ -19,10 +19,11 @@
     ReductionKernelVariables,
 )
 from numba_dpex.core.utils.kernel_launcher import KernelLaunchIRBuilder
+from numba_dpex.dpctl_iface import libsyclinterface_bindings as sycl
```
diptorupd, line 22: `from numba_dpex.core.datamodel.models import dpex_data_model_manager as dpex_dmm` is now unused please remove as part of the PR.

ZzEeKkAa: And `_KernelArgs`

diptorupd: yes

diptorupd:
Few dead code that I noticed after your changes in `parfor_lowerer.py`

- [ ] `_KernelArgs`
- [ ] from numba_dpex.core.datamodel.models import dpex_data_model_manager as dpex_dmm

numba_dpex/core/parfors/parfor_lowerer.py
```diff
@@ -176,45 +108,13 @@ def _submit_parfor_kernel(
 
         local_range = []
```
diptorupd, line 109: A comment is warranted here: "For now the `local_range` is always an empty list as numba_dpex always submits kernels generated for parfor nodes as `range` kernels. The provision is kept here if in future there is newer functionality to submit these kernels as `ndrange`."

numba_dpex/core/parfors/parfor_lowerer.py
```diff
@@ -228,75 +128,61 @@ def _submit_reduction_main_parfor_kernel(
...
-            global_range=global_range,
-            local_range=local_range,
+
+        kl_builder.set_kernel(kernel_ref)
+        kl_builder.set_queue(queue_ref)
+        kl_builder.set_range(global_range, local_range)
+        kl_builder.set_arguments(
+            kernel_fn.kernel_arg_types, kernel_args=kernel_args
         )
+        kl_builder.set_dependant_event_list()
```
diptorupd, line 180: A comment is required: "Kernels generated for `parfor` nodes are always submitted synchronously and do not allow for any dependent events. Thus, the `kl_builder.set_dependant_event_list()` is called without any arguments."

I would have preferred the API forces you to pass in an empty list/tuple. Otherwise the call here can be confusing.

numba_dpex/core/parfors/reduction_helper.py
```diff
@@ -399,7 +399,7 @@ def copy_final_sum_to_host(self, parfor_kernel):
 
         # Create a local variable storing a pointer to a DPCTLSyclQueueRef
         # pointer.
-        curr_queue = ir_builder.get_queue(exec_queue=parfor_kernel.queue)
+        queue_ref = ir_builder.get_queue(exec_queue=parfor_kernel.queue)
```
diptorupd, line 402: Let us stay consistent with naming. `s/ir\_builder/kl\_builder/g`

numba_dpex/core/utils/kernel_launcher.py
```diff
@@ -151,11 +204,11 @@ def build_complex_arg(self, val, ty, arg_list, args_ty_list, arg_num):
         )
         arg_num += 1
 
-    def build_array_arg(
+    def _build_array_arg(  # pylint: disable=too-many-arguments
         self,
         array_val,
         array_data_model,
-        array_rank,
+        array_rank,  # pylint: disable=unused-argument
```
diptorupd, line 211: Why not remove?

numba_dpex/core/parfors/reduction_helper.py
diptorupd: Also, I am assuming the `get_queue` will itself be cleaned up in a follow up PR?

numba_dpex/experimental/launcher.py
```diff
@@ -317,16 +317,15 @@ def codegen(cgctx, builder, sig, llargs):
...
-            local_range_extents=ll_range.local_range_extents,
+        kl_builder = kl.KernelLaunchIRBuilder(kernel_targetctx, builder)
+        kl_builder.set_kernel(kernel_ref)
+        kl_builder.set_queue(queue_ref)
+        kl_builder.set_range(
+            ll_range.global_range_extents, ll_range.local_range_extents
         )
+        kl_builder.set_arguments(ty_kernel_args_tuple, kernel_args)
+        kl_builder.set_dependant_event_list()
+        device_event_ref = kl_builder.submit()
```
diptorupd, line 328: I like the pattern. Much cleaner.

numba_dpex/core/utils/kernel_launcher.py
```diff
@@ -11,6 +17,48 @@
 from numba_dpex.dpctl_iface import libsyclinterface_bindings as sycl
 from numba_dpex.dpctl_iface._helpers import numba_type_to_dpctl_typenum
 
+MAX_SIZE_OF_SYCL_RANGE = 3
```
diptorupd, line 20: There is a place in `parfor_lowerer.py` where the same hard coing is done. It will be goo to move the global to `dpctl_iface` and use it from there.

diptorupd (approved):
Nice clean up. Makes the code readable and reduces redundancies.

ZzEeKkAa:
There is an issue in this change. Before this change parfor_lowerer was using `from numba_dpex.core.datamodel.models import dpex_data_model_manager as dpex_dmm` data model manager, which is the same as kernel data model manager, but now it uses data model manager from context which is equal to numba's data model manager. It is okay as long as number of arguments for USMNdArray and DpnpNdArray is the same, but I would introduce KernelLaunchIRBuilder.kernel_data_model_manager attribute to store the right dmm on instance creation.

numba_dpex/core/utils/kernel_launcher.py
ZzEeKkAa: I've thought it was pending TODO. Removed

numba_dpex/core/parfors/reduction_helper.py
ZzEeKkAa: Yes, it is not part of builder, so it's gonna be removed. Either by getting queue on llvm level or removing this code out of the class.

numba_dpex/core/parfors/parfor_lowerer.py
ZzEeKkAa: Good idea!

PR #1250, opened 2023-12-17
Changes to KernelDispatcher to support numba 0.59

diptorupd:

The `impl_kind` class attribute was removed from numba's `Dispatcher` class in 0.59.0rc1. The `_compiler` attribute for `Dispatcher` is now always set to numba's `_FunctionCompiler` class.

To accommodate for these changes, `numba_dpex.experimental.dispatcher.KernelDispatcher` will conditionally set its `_compiler` attribute to the `_KernelCompiler` class when numba version in use is >0.59, otherwise will default to old behaviour.

- [] Have you added a test, reproducer or referred to an issue with a reproducer?


ZzEeKkAa (approved):
LGTM! Thank you!

PR #1264, opened 2024-01-03
Clean up CMake build

ZzEeKkAa:
Some of the steps in CMake are redundant because they are performed either by python functions in CMake either by scikit build.

I've also figured out that some conda package sets Release cmake target, so to build it with Debug target run:
```bash
CMAKE_ARGS="" CC=icx CXX=icpx python setup.py develop --build-type=Debug
```


diptorupd (approved):
Thank you!

PR #1266, opened 2024-01-04
Feature/switch to dpex target context

ZzEeKkAa:
Enables dpjit target specific context for dpjit target. 

These changes unblocks:
- target specific overloads
- dlpack argument overloads
- simplification of experimental kernel (we can get use of `dpexrt` attribute)

Since this PR switches to dpex specific target context, that means that target specific pipelines pass are applied for all compilations including overloads. That reveals issue with `parfor_legalization_cfd_pass`. Current solution works ok, but does not perform assertion. That happens because we are using undocumented API there and this assertion required to spot those changes in numba's API.

Closes #1027 

Checklist


numba_dpex/core/targets/dpjit_target.py
```diff
@@ -34,31 +34,27 @@ def __init__(self, typingctx, target=DPEX_TARGET_NAME):
...
         self.lower_extensions = {}
+        super().init()
+
         # TODO: initialize nrt once switched to nrt from drt. Most likely we
         # call it somewhere. Double check.
         # https://github.com/IntelPython/numba-dpex/issues/1175
         # Initialize NRT runtime
         # rtsys.initialize(self) # noqa: E800
-        self.refresh()
+        # self.refresh() # noqa: E800
```
diptorupd, line 45: If the `refresh` call is not needed then why not remove instead of comment out?

numba_dpex/decorators.py
```diff
@@ -153,22 +154,19 @@ def dpjit(*args, **kws):
         warnings.warn(
             "pipeline class is set for dpjit and is ignored", RuntimeWarning
         )
-        del kws["forceobj"]
+        del kws["pipeline_class"]
```
diptorupd, line 157: :smiley_cat:

numba_dpex/core/targets/dpjit_target.py
ZzEeKkAa: I was unsure if it is needed for nrt. If not - I can just remove it.

diptorupd: Its purpose was to load any `dpjit`-specific function registries when the target is initialized. I see that the `refresh` function is removed as well. Let us remove, we will revisit when we have `dpjit`-specific registries.

diptorupd (approved):
great work to tie these loose ends and move the dpjit target work towards completion. Thank you!

Issue #1010, opened 2023-04-20
Numba_dpex breaks numba execution


ZzEeKkAa:
The following snippet stops working in `import numba_dpex` is added. We should prevent overriding vanilla numba behavior.

```python
import numba
import numpy
import numba_dpex 

@numba.njit
def foo(a):
   return a + 1

a = numpy.asarray([1j])
print(foo(a))
```

The issue was spotted while working on `npbench` and `dpbench` integration. Thank you @AlexanderKalistratov for providing small reproducer.

ZzEeKkAa:
Issue is happening because `numba_dpex` is doing monkey patch of math operators (in particular for operator `add`) in `numba_dpex.dpnp_iface.dpnp_ufunc_db.py` in `_fill_ufunc_db_with_dpnp_ufuncs`. That results that from `numba.np.ufunc_db import _ufunc_db` is getting overwritten and no longer supports original signature.

diptorupd:
One more reproducer submitted by @Hardcode84 

```Python
import numba_dpex as nbm

import numba as nb
import numpy as np

def foo(a, b):
    return np.add(a, b)

foo1 = nb.njit(foo)

a = np.ones((10,),dtype=np.complex128)
print(foo(a, 1.5))
print(foo1(a, 1.5))


[2.5+0.j 2.5+0.j 2.5+0.j 2.5+0.j 2.5+0.j 2.5+0.j 2.5+0.j 2.5+0.j 2.5+0.j
 2.5+0.j]
Traceback (most recent call last):
  File "D:\projs\dpcomp\test-dpex-break.py", line 13, in <module>
    print(foo1(a, 1.5))
  File "C:\Users\Vano\miniconda3\envs\dpex-env\lib\site-packages\numba\core\dispatcher.py", line 468, in _compile_for_args
    error_rewrite(e, 'typing')
  File "C:\Users\Vano\miniconda3\envs\dpex-env\lib\site-packages\numba\core\dispatcher.py", line 409, in error_rewrite
    raise e.with_traceback(None)
numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
No implementation of function Function(<ufunc 'add'>) found for signature:

 >>> add(array(complex128, 1d, C), float64)

There are 2 candidate implementations:
  - Of which 2 did not match due to:
  Overload in function 'Numpy_rules_ufunc.generic': File: numba\core\typing\npydecl.py: Line 97.
[...]

Issue #178, opened 2021-01-20
Add support for all operators in kernels


1e-to:
- [x] operator.add() (+)
- [ ] operator.and_() (&)
- [x] operator.eq() (==)
- [x] operator.floordiv() (//)
- [x] operator.ge() (>=)
- [x] operator.gt() (>)
- [x] operator.iadd() (+=)
- [ ] operator.iand() (&=)
- [x] operator.ifloordiv() (//=)
- [ ] operator.ilshift() (<<=)
- [x] operator.imod() (%=)
- [x] operator.imul() (*=)
- [ ] operator.invert() (~=)
- [ ] operator.ior() (|=)
- [x] operator.ipow() (**=)
- [ ] operator.irshift() (>>=)
- [x] operator.isub() (-=)
- [x] operator.itruediv() (/=)
- [ ] operator.ixor() (^=)
- [x] operator.le() (<=)
- [x] operator.lshift() (<<)
- [x] operator.lt() (<)
- [x] operator.mod() (%)
- [x] operator.mul() (*)
- [x] operator.ne() (!=)
- [x] operator.neg() (- x)
- [x] operator.not_() (not x)
- [x] operator.or_() (or)
- [x] operator.pos() (+ x)
- [x] operator.pow() (**)
- [ ] operator.rshift() (>>)
- [x] operator.sub() (-)
- [x] operator.truediv() (/)
- [ ] operator.xor() (^)

PR #1267, opened 2024-01-04
Create dpnp specific ufunc db

ZzEeKkAa:
Add tests to ensure that `numba.njit` works with `numpy` after importing `numba_dpex`. Aka lazy testing if we break `numba`'s default behavior.

Fixes: #1010 


diptorupd (approved):
Thanks you! Another one bites the dust!

diptorupd (changes requested):
We have some CI failures that should be investigated.

diptorupd (approved):
We should revisit if we should use deepcopy when copying the numba ufunc_db so as to not have a special case for missing operators. Can be revisited after we reevaluate #178

PR #1269, opened 2024-01-05
Make overloads and intrinsic target specific

ZzEeKkAa:
Make overloads and intrinsic target specific, so we don't accidentally break numba.


diptorupd:
Overall the changes make sense and I am happy that we finally got it to work. We need some test cases though. 

a) Something like calling `dpnp.empty` inside `numba.njit` should raise some error from Numba about unrecognized function.
b) Checking the function definitions for dpjit target context to see if these functions are available.
c) A negative test that these functions are not available in the `BaseContext` function definitions.

Also, do note that `intrinsics` ignore the target keyword so they will be available in `BaseContext`.

diptorupd (approved):
Thank you!

PR #1272, opened 2024-01-09
Allow importing numba-dpex even if no SYCL device is detected by dpctl

### LINKED ISSUE #304: Allow importing numba-dpex even if dpctl shows that there are no SYCL platforms
Currently `import numba-dppy` raises exception if SYCL platform is not available.
https://github.com/IntelPython/numba-dppy/blob/6033fb832444e8e9fb5a0e12d300eaf4fdf438f9/numba_dppy/__init__.py#L527

It uses dpctl:
https://github.com/IntelPython/numba-dppy/blob/6033fb832444e8e9fb5a0e12d300eaf4fdf438f9/numba_dppy/config.py#L21

It means that if SYCL is not available we can not import `numba-dppy` and its tests at all.

Why not we able to import `numba-dppy` and run tests which do not use SYCL devices? We could skip all SYCL dependent tests but others could be run.

@diptorupd @reazulhoque 
What do you think? Is it not too strict to prohibit import `numba-dppy`?

It is related to #38. It is confusing when installed package behaves like it is not installed.

diptorupd:

Numba-dpex checks at module importation that at least one non-host SYCL device is available on the system. The check leads to the following `ImportError` if we try to import numba-dpex on a system that has no such device:

```bash
Traceback (most recent call last):
  File "/localdisk/work/diptorup/devel/numba-dpex/numba_dpex/config.py", line 40, in <module>
    HAS_NON_HOST_DEVICE = dpctl.select_default_device()
  File "dpctl/_sycl_device_factory.pyx", line 320, in dpctl._sycl_device_factory.select_default_device
  File "dpctl/_sycl_device_factory.pyx", line 335, in dpctl._sycl_device_factory.select_default_device
dpctl._sycl_device.SyclDeviceCreationError: Default device is unavailable.
Traceback (most recent call last):
  File "/localdisk/work/diptorup/devel/numba-dpex/numba_dpex/examples/kernel/vector_sum.py", line 8, in <module>
    import numba_dpex as ndpx
  File "/localdisk/work/diptorup/devel/numba-dpex/numba_dpex/__init__.py", line 114, in <module>
    if config.HAS_NON_HOST_DEVICE:
  File "/localdisk/work/diptorup/devel/numba-dpex/numba_dpex/config.py", line 33, in __getattr__
    return getattr(config, name)
AttributeError: module 'numba.core.config' has no attribute 'HAS_NON_HOST_DEVICE'
```
The check is a legacy of the time when numba-dpex needed to access a default queue to determine where a kernel should be launched if the arguments to the kernel where NumPy arrays. Since numba-dpex does not any longer support launching kernels with NumPy ndarray arguments and all offload target decisions are based on the compute follows data programming model, the check can be removed.

The PR removes the check. Without the check a user will get the exception at the array creation stage from the array library (dpnp or dpctl.tensor). The behavior is more intuitive to an user and correctly points out the reason for the exception.

```bash
Traceback (most recent call last):
  File "/localdisk/work/diptorup/devel/numba-dpex/numba_dpex/examples/kernel/vector_sum.py", line 45, in <module>
    main()
  File "/localdisk/work/diptorup/devel/numba-dpex/numba_dpex/examples/kernel/vector_sum.py", line 34, in main
    a = dpnp.random.random(N)
  File "/nfs/site/home/diptorup/.conda/envs/dpex-devel/lib/python3.10/site-packages/dpnp-0.12.0-py3.10-linux-x86_64.egg/dpnp/random/dpnp_iface_random.py", line 1214, in random
    return random_sample(size=size, device=device, usm_type=usm_type, sycl_queue=sycl_queue)
  File "/nfs/site/home/diptorup/.conda/envs/dpex-devel/lib/python3.10/site-packages/dpnp-0.12.0-py3.10-linux-x86_64.egg/dpnp/random/dpnp_iface_random.py", line 1316, in random_sample
    rs = _get_random_state(device=device, sycl_queue=sycl_queue)
  File "/nfs/site/home/diptorup/.conda/envs/dpex-devel/lib/python3.10/site-packages/dpnp-0.12.0-py3.10-linux-x86_64.egg/dpnp/random/dpnp_iface_random.py", line 106, in _get_random_state
    sycl_queue = dpnp.get_normalized_queue_device(device=device, sycl_queue=sycl_queue)
  File "/nfs/site/home/diptorup/.conda/envs/dpex-devel/lib/python3.10/site-packages/dpnp-0.12.0-py3.10-linux-x86_64.egg/dpnp/dpnp_iface.py", line 375, in get_normalized_queue_device
    return dpt._device.normalize_queue_device(sycl_queue=sycl_queue, device=device)
  File "/nfs/site/home/diptorup/.conda/envs/dpex-devel/lib/python3.10/site-packages/dpctl/tensor/_device.py", line 171, in normalize_queue_device
    d = Device.create_device(d)
  File "/nfs/site/home/diptorup/.conda/envs/dpex-devel/lib/python3.10/site-packages/dpctl/tensor/_device.py", line 76, in create_device
    _dev = dpctl.SyclDevice()
  File "dpctl/_sycl_device.pyx", line 348, in dpctl._sycl_device.SyclDevice.__cinit__
dpctl._sycl_device.SyclDeviceCreationError: Could not create a SyclDevice from default selector
```

The test for this change is not easily testable using pytest. Instead a new smoke test was added to the Github conda-package workflow.

Fixes #304


numba_dpex/__init__.py
```diff
@@ -111,33 +111,30 @@ def parse_sem_version(version_string: str) -> Tuple[int, int, int]:
...
+    get_local_size,
+    get_num_groups,
+    get_work_dim,
+    local,
+    mem_fence,
+    private,
+    sub_group_barrier,
+)
+
+DEFAULT_LOCAL_SIZE = []
```
chudur-budur, line 134: Why do we still have this? We need to completely nuke the old way of setting the kernel lauch parameters.

chudur-budur (approved):
LGTM!

numba_dpex/__init__.py
diptorupd: Right. `DEFAULT_LOCAL_SIZE` is referenced in the vectorizers.py module. The vectorizers.py is badly broken and non functional at the moment that is why I did not remove the code from `__init__.py`. I am going to do that in a follow up PR.

diptorupd: Created #1273 to keep track of the pending clean up.

PR #1276, opened 2024-01-10
Update license and readme to fix wheels metadata

ekomarova:

This PR is designed to fix metadata in wheels. `License :: OSI Approved :: Apache 2.0` is actually incorrect definition for a license in wheels, because of this, when uploading a package to pypi, we have to repackage wheel by changing the license to the correct one. Also, there are no instructions in README for installing wheels and conda packages


diptorupd (approved):
Thank you!

PR #1278, opened 2024-01-10
Pin dpcpp used at build time to bypass broken package.

diptorupd:
Update the conda recipe meta.yaml to use same dpcpp pinning as dpctl.


diptorupd:
Did not help

Issue #146, opened 2021-01-12
Disallow numpy calls inside numba_dpex.kernel decorated functions


diptorupd:
NumPy calls should not be usable inside a `numba_dpex.kernel` decorated function. We currently support some calls such as `np.sin`, but this is an anti-pattern and only `math` functions should be supported inside the `dppy.kernel` decorator.

PR #1282, opened 2024-01-18
Feature/use dpnp specific registry

ZzEeKkAa:
Provide dpnp specific registry so we don't override numba's default behaviour

### DIFF
--- a/numba_dpex/core/targets/dpjit_target.py
+++ b/numba_dpex/core/targets/dpjit_target.py
@@ -7,6 +7,7 @@
 
 from functools import cached_property
 
+from numba.core import typing
 from numba.core.compiler_lock import global_compiler_lock
 from numba.core.cpu import CPUContext
 from numba.core.imputils import Registry
@@ -26,6 +27,10 @@ DPEX_TARGET_NAME = "dpex"
 target_registry[DPEX_TARGET_NAME] = Dpex
 
 dpex_function_registry = Registry()
+
+
+class DpexTypingContext(typing.Context):
+    """Custom typing context to support dpjit compilation."""
 
 
 class DpexTargetContext(CPUContext):
--- a/numba_dpex/core/targets/dpjit_target.py
+++ b/numba_dpex/core/targets/dpjit_target.py
@@ -32,6 +32,13 @@ dpex_function_registry = Registry()
 class DpexTypingContext(typing.Context):
     """Custom typing context to support dpjit compilation."""
 
+    def load_additional_registries(self):
+        """Register dpjit specific functions like dpnp ufuncs."""
+        from numba_dpex.core.typing import dpnpdecl
+
+        self.install_registry(dpnpdecl.registry)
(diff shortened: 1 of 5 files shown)


numba_dpex/core/targets/dpjit_target.py
```diff
@@ -28,6 +29,17 @@ class Dpex(CPU):
 dpex_function_registry = Registry()
 
 
+class DpexTypingContext(typing.Context):
+    """Custom typing context to support dpjit compilation."""
+
+    def load_additional_registries(self):
+        """Register the OpenCL API and math and other functions."""
```
diptorupd, line 36: Incorrect docstring?

numba_dpex/core/targets/kernel_target.py
```diff
@@ -111,12 +111,15 @@ def load_additional_registries(self):
...
         from numba.core.typing import cmathdecl, enumdecl, npydecl
 
+        from numba_dpex.core.typing import dpnpdecl
+
         from ...ocl import mathdecl, ocldecl
 
         self.install_registry(ocldecl.registry)
         self.install_registry(mathdecl.registry)
         self.install_registry(cmathdecl.registry)
         self.install_registry(npydecl.registry)
```
diptorupd, line 121: Do we still need `npydecl` to be registered?

diptorupd: If we are able to get rid of `npydecl` it should resolve #146. Whether we should allow `dpnp` calls in kernel is another matter.

numba_dpex/core/targets/dpjit_target.py
diptorupd: The above comment is in the context of "OpenCL API", `dpjit` neither supports the functions in ocldecl registry, nor shout it.

numba_dpex/core/targets/kernel_target.py
ZzEeKkAa: Apparently #146 was resolved in #1267 . It broken dpbench workflow that was calling numpy function in kernel

numba_dpex/core/targets/dpjit_target.py
ZzEeKkAa: Yes, copy paste from kernel

numba_dpex/core/targets/kernel_target.py
ZzEeKkAa: We cannot get rid of because of the `truediv` workaround. We need to investigate it deeper to be able generate dpnp specific ufunc without calling `numpy.truediv` function. It is out of scope of this PR, so no changes needed.

diptorupd: Thanks for the explanation. Can you add a comment with the explanation on the line where the `npydecl.registry` is installed?

diptorupd (approved):
Barring the small nit about the comment, LGTM!

PR #1283, opened 2024-01-19
Improve built-in ops and dpnp universal functions testing

ZzEeKkAa:
Add full coverage on ufuncs and built-in operators. Motivated by https://github.com/IntelPython/numba-dpex/issues/178

It appears that we were not aware that these functions are not supported:
- "cbrt"

We also does not support (may be we need proper input data):
- `matmul` operator (`a @ b`)
- `imatmul` operator (`a @= b`)
- `not_` operator (`not a`) (because numpy does not support it)

Open question:
- Does it mean if the operator works in parfor, than it will work in kernel?

Checklist:


diptorupd:
@ZzEeKkAa Thank you for the comprehensive evaluation. I am replying to some of your questions and comments in the PR description.

1.  > It appears that we were not aware that these functions are not supported:
    > "cbrt"
    > "mod"

Please log separate issues to track the gap.

2.  > We also does not support (may be we need proper input data):
     >
     >   matmul operator (a @ b)
     >   imatmul operator (a @= b)
     >   not_ operator (not a)

The matmul operators not being supported is expected. I believe the `matmul` operators are not treated as parfor and Numba lowers them as library calls. Numba-dpex used to have the same functionality at one point.

The `not_` not being supported needs investigation. Please open an issue so that we can keep track of it. 

3.  > Does it mean if the operator works in parfor, than it will work in kernel?

That is a correct assumption. However, we should note a subtle difference. There are various types of expressions that can be lowered to `parfor` nodes: numpy calls on arrays and built-in operations on arrays (both of which are vector operations), numpy calls and built in operations on scalars inside a `prange` loop (the operation here is a scalar op). In a kernel, only the scalar versions of all these operators and ufuncs can be used.

It is a good idea to add similar unit tests for the `kernel` decorator as well. We will be explicitly verifying that lowering a scalar operation works.

numba_dpex/tests/dpjit_tests/parfors/test_builtin_ops.py
```diff
@@ -81,13 +365,16 @@ def test_built_in_operators(shape, dtype, usm_type, func):
     else:
         assert c.shape == shape
 
-    if func != parfor_divide:
+    if func not in {parfor_truediv} | comparison_funcs:
```
diptorupd, line 368: Did you mean to use bitwise or?

numba_dpex/core/typing/dpnpdecl.py
```diff
@@ -179,12 +179,16 @@ def install_operations(cls):
         "ldexp",
         "spacing",
         "isnat",
+        "cbrt",
```
diptorupd, line 182: Please split out the changes related to adding `cbrt` to unsupported list and removing `mod` from _aliases into a separate commit with meaningful commit message. This commit from description indicates only test cases were changes.

numba_dpex/tests/dpjit_tests/parfors/test_builtin_ops.py
ZzEeKkAa: yes, it is set sum

diptorupd: 👍 I did know such a thing existed

diptorupd (approved):
nice changes. Making unit testing more comprehensive is always welcome!

PR #1268, opened 2024-01-05
Fix dpnp n-d array strides issue in kernel

chudur-budur:
This PR will fix the issue #572.

- [x] Have you provided a meaningful PR description?
- [x] Have you added a test, reproducer or referred to an issue with a reproducer?
- [x] Have you tested your changes locally for CPU and GPU devices?
- [x] Have you made sure that new changes do not introduce compiler warnings?
- [x] If this PR is a work in progress, are you filing the PR as a draft?

Issue #363, opened 2021-05-14
Do we envision supporting strides and non-C contiguous data layouts?


PokhodenkoSA:
Do we envision supporting strides and non-C contiguous data layouts?

_Originally posted by @diptorupd in https://github.com/IntelPython/numba-dppy/pull/287#r617173732_

PokhodenkoSA:
@diptorupd
What ideas do you have about it?

diptorupd:
@PokhodenkoSA Numba supports strides and `dpctl.tensor.usm_ndarray` supports strides. I plan to work to add support for strides in `kernel` first and then `njit`. However, the changes for strides will happen post 2021.3.

diptorupd:
Strides and non-C contiguous layout now work as expected. The new feature has been verified by unit tests added on #1285

PR #1285, opened 2024-01-20
Refactor all tests for strided dpnp arrays in kernel with different layouts

### LINKED ISSUE #572: numba_dpex's kernel disregards strides of input array
```python
import dpctl
import numba_dpex as dppy
import numpy as np
import dpnp

import dpctl.tensor as dpt


@dppy.kernel
def data_parallel_sum(a, b, c):
    """
    Vector addition using the ``kernel`` decorator.
    """
    i = dppy.get_global_id(0)
    c[i] = a[i] + b[i]


def driver(a, b, c, global_size):
    print("A : ", (a.shape, a.strides))
    print("B : ", (b.shape, b.strides))
    data_parallel_sum[dppy.Range(global_size)](a, b, c)
    print("C ", (c.shape, c.strides))
    # assert np.allclose(dpt.asnumpy(c), dpt.asnumpy(a) + dpt.asnumpy(b))


global_size = 10
N = global_size
print("N", N)

a = dpnp.random.random(N)  # C-contig
b = dpnp.random.random(N)  # C-contig
out = dpnp.zeros((2 * N,), dtype=a.dtype)  # C-contig
c = out[::2]  # Strided, every other element

driver(a, b, c, global_size)

print(c)
print(out)

```

Executing this script outputs

```
N 10
Using device ...
    Name            Intel(R) UHD Graphics [0x9bca]
    Driver version  1.1.20678
    Vendor          Intel(R) Corporation
    Profile         FULL_PROFILE
    Filter string   level_zero:gpu:0
A :  ((10,), (1,))
B :  ((10,), (1,))
C  ((10,), (2,))
Done...
[0.83166494 1.27176889 0.61416447 0.57984313 1.04820722 0.
 0.         0.         0.         0.        ]
[0.83166494 0.50883122 1.27176889 1.08111028 0.61416447 0.77052362
 0.57984313 1.55592601 1.04820722 1.12209132 0.         0.
 0.         0.         0.         0.         0.         0.
 0.         0.        ]
```

The expected result should have been

```
[0.83166494 0.50883122 1.27176889 1.08111028 0.61416447 0.77052362
 0.57984313 1.55592601 1.04820722 1.12209132]
[0.83166494 0.       0.50883122  0.       1.27176889  0.        1.08111028  0.        
0.61416447 0.        0.77052362  0.       0.57984313  0.        1.55592601  0.    
1.04820722  0.       1.12209132    0.  ]
```

**UPDATE** : Updated the reproducer to work with latest numba-dpex.

chudur-budur:
1. Strided array correctly works with the current code.
2. This PR performs unit test for all situations (differing dimensions, shape, strides, view-of-views etc.)
3. This PR performs unit test for both `C` and `F` contiguous arrays.
4. This is relevant to the experimental kernel, will not work with the old kernel code.

This PR finally resolves #572 and supersedes #1268 and #1271

Fixes #572, #363

### DIFF
--- a/numba_dpex/tests/experimental/test_strided_dpnp_array_in_kernel.py
+++ b/numba_dpex/tests/experimental/test_strided_dpnp_array_in_kernel.py
@@ -2,23 +2,72 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import math
+
 import dpnp
+import numpy as np
+import pytest
 
 import numba_dpex
 import numba_dpex.experimental as exp_dpex
-from numba_dpex import NdRange, Range
+from numba_dpex import Range
+
+
+def get_order(a):
+    """Get order of an array.
+
+    Args:
+        a (numpy.ndarray, dpnp.ndarray): Input array.
+
+    Raises:
+        Exception: _description_
+
+    Returns:
+        str: 'C' if c-contiguous, 'F' if f-contiguous or 'A' if aligned.
+    """
+    if a.flags.c_contiguous and not a.flags.f_contiguous:
+        return "C"
+    elif not a.flags.c_contiguous and a.flags.f_contiguous:
+        return "F"
+    elif a.flags.c_contiguous and a.flags.f_contiguous:
+        return "A"
+    else:
+        raise Exception("Unknown order/layout")
 
 
 @exp_dpex.kernel
 def change_values_1d(x, v):
+    """Assign values in a 1d dpnp.ndarray
+
+    Args:
+        x (dpnp.ndarray): Input array.
+        v (int): Value to be assigned.
+    """
     i = numba_dpex.get_global_id(0)
     p = x[i]  # getitem
     p = v
     x[i] = p  # setitem
 
 
+def change_values_1d_func(a, p):
+    """Assign values in a 1d numpy.ndarray
+
+    Args:
+        a (numpy.ndarray): Input array.
+        p (int): Value to be assigned.
+    """
+    for i in range(a.shape[0]):
+        a[i] = p
+
+
 @exp_dpex.kernel
 def change_values_2d(x, v):
+    """Assign values in a 2d dpnp.ndarray
+
+    Args:
+        x (dpnp.ndarray): Input array.
+        v (int): Value to be assigned.
+    """
     i = numba_dpex.get_global_id(0)
     j = numba_dpex.get_global_id(1)
     p = x[i, j]  # getitem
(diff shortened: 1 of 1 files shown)


numba_dpex/tests/experimental/test_strided_dpnp_array_in_kernel.py
```diff
@@ -2,32 +2,99 @@
...
 @exp_dpex.kernel
 def change_values_1d(x, v):
+    """Assign values in a 1d dpnp.ndarray
+
+    Args:
+        x (dpnp.ndarray): Input array.
+        v (int): Value to be assigned.
+    """
     i = numba_dpex.get_global_id(0)
     p = x[i]  # getitem
```
diptorupd, line 47: What is the point of the `p = x[i]`? Why not do `x[i] = v`?

numba_dpex/tests/experimental/test_strided_dpnp_array_in_kernel.py
```diff
@@ -36,74 +103,192 @@ def change_values_3d(x, v):
...
+        a (numpy.ndarray): Input array.
+        p (int): Value to be assigned.
+    """
+    for i in range(a.shape[0]):
+        for j in range(a.shape[1]):
+            for k in range(a.shape[2]):
+                a[i, j, k] = p
+
+
+@pytest.mark.parametrize("N", [8, 16, 32, 64, 128, 256, 512])
```
diptorupd, line 119: Why do we need so many `N` values? I do not feel they add much to the test coverage and only make the suite run longer.

diptorupd:
The logic in `_dpexrt_python.c::DPEXRT_sycl_usm_ndarray_from_python` does indeed correctly compute strides for `F` and `C` order. It is because of a quirk in the dpctl implementation of stride computation. dpctl C-API returns no stride information when the fastest changing dimension has a unit stride, _i.e._, `C` order layout. In other cases that includes strided views and `F` order, the actual stride value is returned. Our stride conversion logic, from dpctl's element count to Numba/NumPy's byte count, in  `_dpexrt_python.c::DPEXRT_sycl_usm_ndarray_from_python` works for this reason. When a stride vector is received from dpctl we correctly convert them to number of bytes and for the default `C` order case strides are calculated as number of bytes.

Since, `experimental.kernel` now uses Numba's unbox/box routines the default array indexing for `__getitem__` and `__setitem__` calls on `dpnp.ndarray` or `dpctl.usm_ndarray` works out of the box.

An example to visually verify the new behavior:

```python
import numba_dpex as dpex
import dpnp

import numpy

@dpex.dpjit
def foo(a, b):
    print(a.strides)
    print(b.strides)
    return a, b

a = numpy.ones(24, order="F").reshape(4,6, order="F")
da = dpnp.ones(24, order="F").reshape(4,6, order="F")

x = numpy.reshape(numpy.arange(5*6*7*8), (5,6,7,8)).transpose(2,3,1,0)
dx = dpnp.reshape(dpnp.arange(5*6*7*8), (5,6,7,8)).transpose(2,3,1,0)

print(a.strides)
print(da.strides)

print(x.strides)
print(dx.strides)

x, dx = foo(x, dx)
a, da = foo(a, da)

print(a.strides)
print(da.strides)

print(x.strides)
print(dx.strides)

```

diptorupd:
I have updated the misleading FIXME in _dpexrt_python.c` in https://github.com/IntelPython/numba-dpex/pull/1286

numba_dpex/tests/experimental/test_strided_dpnp_array_in_kernel.py
chudur-budur: fixed.

diptorupd (approved):
Thank you!

PR #1286, opened 2024-01-21
Update FIXME comment in stride computation.

diptorupd:
Updates the `FIXME` comment in `_dpexrt_python.c`. The comment was incorrect as the logic to compute strides from dpctl/dpnp actually works for all layouts and the `experimental` kernel now can support `F` and strided layouts.

### DIFF
--- a/numba_dpex/core/runtime/_dpexrt_python.c
+++ b/numba_dpex/core/runtime/_dpexrt_python.c
@@ -866,13 +866,14 @@ static int DPEXRT_sycl_usm_ndarray_from_python(PyObject *obj,
     for (i = 0; i < ndim; ++i, ++p)
         *p = shape[i];
 
-    // DPCTL returns a NULL pointer if the array is contiguous. dpctl stores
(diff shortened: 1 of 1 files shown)


numba_dpex/core/runtime/_dpexrt_python.c
```diff
@@ -866,13 +866,14 @@ static int DPEXRT_sycl_usm_ndarray_from_python(PyObject *obj,
...
         *p = shape[i];
 
-    // DPCTL returns a NULL pointer if the array is contiguous. dpctl stores
-    // strides as number of elements and Numba stores strides as bytes, for
-    // that reason we are multiplying stride by itemsize when unboxing the
-    // external array.
-
-    // FIXME: Stride computation should check order and adjust how strides are
-    // calculated. Right now strides are assuming that order is C contigous.
+    // dpctl returns a NULL pointer for the stride vector if the array has a
```
chudur-budur, line 869: Yup, this is better than previous comment.

chudur-budur (approved):
LGTM!

PR #1289, opened 2024-01-22
Update the comments around why we reverse dpex.Range indices in kernel_launcher.py

diptorupd:


ZzEeKkAa (approved):
LGTM! Not sure if it is reproduceable only on my machine, but most tests crash for `opencl:gpu` devices

Issue #1270, opened 2024-01-05
Call to divide ufunc generates parfor with numpy divide call instead of / operator


ZzEeKkAa:
The issue may be due to the fact that there are two divide operations `numpy.divide` and `numpy.true_divide`. Both for `numpy` and `dpnp` they are using the same realization, so no sure what is the purpose for it. It is also may be the case that `dpnp.true_divide` calls `dpnp.divide`, but numpy just refer both function to the same location.

Generated parfor for divide operation:

```
    $np_g_var.20 = global(np: <module 'numpy' from '/home/yevhenii/.miniconda3/envs/dpex-dev/lib/python3.11/site-packages/numpy/__init__.py'>) ['$np_g_var.20']
    $div_attr.21 = getattr(value=$np_g_var.20, attr=divide) ['$div_attr.21', '$np_g_var.20']
    $expr_out_var.17 = call $div_attr.21($arg_out_var.18, $arg_out_var.19, func=$div_attr.21, args=[Var($arg_out_var.18, test_builtin_ops.py:27), Var($arg_out_var.19, test_builtin_ops.py:27)], kws=(), vararg=None, varkwarg=None, target=None) ['$arg_out_var.18', '$arg_out_var.19', '$div_attr.21', '$expr_out_var.17']
```

And generated parfor for multiply operation:
```
    $expr_out_var.17 = $arg_out_var.18 * $arg_out_var.19 ['$arg_out_var.18', '$arg_out_var.19', '$expr_out_var.17']
```

So the expected instruction should be something like:
```
    $expr_out_var.17 = $arg_out_var.18 / $arg_out_var.19 ['$arg_out_var.18', '$arg_out_var.19', '$expr_out_var.17']
```

Because of this we need to include dpnp.divide ufunc to the `_dpnp_ufunc_db`.

Blocks: https://github.com/IntelPython/numba-dpex/issues/146

PR #1290, opened 2024-01-22
Feature/use dpnp specific parfor

ZzEeKkAa:
Introduce dpjit/dpnp specific pass. Instead of patching numba function update all call tree to generate right call. This unblocked setting dpnp specific instructions.

Checklist:

Fixes: #1270
Fixes: #146


numba_dpex/core/parfors/parfor_pass.py
```diff
@@ -0,0 +1,214 @@
+import warnings
```
diptorupd, line 1: License header is missing

diptorupd:
@ZzEeKkAa overall the approach is good. I suggest you explore one thing. Instead of completely duplicating the upstream `ParforPass` and using our version of the upstream pass, maybe we create a supplementary pass. Our pass runs after the upstream pass and does the extra analysis that is needed to identify as `parfor` nodes dpnp array expressions that are missed by the upstream pass. The benefit of doing so, if we are able to do it that is, will be lesser maintenance overhead on our part.

diptorupd:
> Our pass runs after the upstream pass and does the extra analysis that is needed to identify as `parfor` nodes dpnp array expressions that are missed by the upstream pass.

That is just do`ConvertDPNPPass(self).run(self.func_ir.blocks)`

PR #1292, opened 2024-01-23
Remove spirv-tool call generators from spirv_generator.py.

diptorupd:

    - The spirv-tool package provides various utilities to
      disassemble, validate and potentially optimize a
      SPIR-V binary file. Numba-dpex no longer has spirv-tools
      as a dependency and it is up to the user if they want to
      try out the funtionalities provided by spirv-tools.

      Having the option to call spirv-tool utilities from
      spirv_generator is unused and potentially confusing. The
      commit removes all such options and renames the helper
      class from CmdLine to _SprivGenerator.

    - Also removed the config option to run spirv validator.


ZzEeKkAa (approved):
LGTM!

PR #1300, opened 2024-01-30
Implementation of sub-group barrier operation

adarshyoga:

This PR contains implementation of sub_group_barrier function that synchronizes across all work-items in a sub-group. It currently lacks a test case to verify the functionality of the barrier call. The tests need to be implemented once sub-group feature is added.

### DIFF
--- a/numba_dpex/experimental/__init__.py
+++ b/numba_dpex/experimental/__init__.py
@@ -8,7 +8,10 @@ yet production ready.
 
 from numba.core.imputils import Registry
 
-from ._kernel_dpcpp_spirv_overloads import _atomic_ref_overloads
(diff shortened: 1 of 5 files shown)


ZzEeKkAa:
After https://github.com/IntelPython/numba-dpex/pull/1310 is merged we have to follow sycl syntax and make it work using `group_barrier(nd_item.get_sub_group(), MemoryScope.DEVICE)`

adarshyoga:
Now that the `group_barrier` function accepts a reference to a `group` as argument, we no longer need a separate `sub_group_barrier`.
A barrier across a `sub-group` can be achieved by calling `group_barrier` and passing a reference to a `sub-group` which will extend `group`.

PR #1307, opened 2024-02-02
Make numba 0.58 the minimum supported version

diptorupd:

Now that Numba 0.59 is out, we can deprecate support for Numba 0.57 and remove one of the numba monkey patches we have been carrying along for a while. The patch was pushed upstream and is available since 0.58 release.

Update all configs and remove the patch.


ZzEeKkAa (approved):
LGTM!

PR #1312, opened 2024-02-03
Atomic compare_exchange implementation

adarshyoga:

This PR contains the implementation of compare_exchange_strong and compare_exchange_weak atomic functions. It contains a test case that verifies both success and failure scenarios.

### DIFF
--- a/numba_dpex/experimental/_kernel_dpcpp_spirv_overloads/_atomic_ref_overloads.py
+++ b/numba_dpex/experimental/_kernel_dpcpp_spirv_overloads/_atomic_ref_overloads.py
@@ -34,6 +34,7 @@ from ._spv_atomic_inst_helper import (
 )
 from .spv_fn_generator import (
     get_or_insert_atomic_load_fn,
+    get_or_insert_spv_atomic_compare_exchange_fn,
     get_or_insert_spv_atomic_exchange_fn,
     get_or_insert_spv_atomic_store_fn,
 )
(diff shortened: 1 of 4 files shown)


numba_dpex/kernel_api/atomic_ref.py
```diff
@@ -200,3 +200,55 @@ def exchange(self, val):
         old = self._ref[self._index]
         self._ref[self._index] = val
         return old
+
+    def compare_exchange_weak(self, expected, desired, expected_idx=0):
```
diptorupd, line 204: I get the rationale, and having discussed it with you I is clear to me that `expected` needs to be an Array type. However, I still feel that from a general usage purpose it is rather confusing. We should create a `Ref` type in numba-dpex to make the usage clear.

numba_dpex/kernel_api/atomic_ref.py
```diff
@@ -200,3 +200,55 @@ def exchange(self, val):
...
+        """Compares the value of the object referenced by the AtomicRef
+        against the value of ``expected[expected_idx]``.
+        If the values are equal, attempts to replace the value of the
+        referenced object with the value of ``desired``.
+        Otherwise assigns the original value of the
+        referenced object to ``expected[expected_idx]``.
+
+        Args:
+            expected : Vector containing the expected value of the
```
diptorupd, line 213: Python has no type called `Vector`. Perhaps array is the right term. But, again the usage will be confusing as to why someone needs to create an array to essentially pass a single element.

numba_dpex/kernel_api/atomic_ref.py
```diff
@@ -200,3 +200,55 @@ def exchange(self, val):
...
+        replacement operation were successful.
+
+        """
+        if self._ref[self._index] == expected[expected_idx]:
+            self._ref[self._index] = desired
+            return True
+        expected[expected_idx] = self._ref[self._index]
+        return False
+
+    def compare_exchange_strong(self, expected, desired, expected_idx=0):
```
diptorupd, line 230: What is the difference between the two variants of `CAS`? When should one use one over the other?

diptorupd: Perhaps we should just have one variant?

numba_dpex/kernel_api/atomic_ref.py
adarshyoga: Created #1316 for tracking work on new array ref type, which would be handled in a separate PR.

numba_dpex/experimental/_kernel_dpcpp_spirv_overloads/_atomic_ref_overloads.py
```diff
@@ -323,6 +324,108 @@ def _intrinsic_exchange_gen(context, builder, sig, args):
...
+        desired_arg = args[2]
+        atomic_ref_ptr = builder.extract_value(
+            args[0],
+            context.data_model_manager.lookup(sig.args[0]).get_field_position(
+                "ref"
+            ),
+        )
+        # add conditional bitcast for atomic_ref pointer,
+        # expected[expected_idx], and desired
+        if sig.args[0].dtype == types.float32:
```
diptorupd, line 364: We can do it like follows as well to reduce code duplication:

```python
if isinstance(sig.args[0].dtype, types.Float):
    bitwidth = sig.args[0].dtype.bitwidth
    atomic_ref_ptr = builder.bitcast(
                atomic_ref_ptr,
                llvmir.PointerType(
                    llvmir.IntType(bitwidth), addrspace=sig.args[0].address_space
                ),
            )
            expected_arg = builder.bitcast(expected_arg, llvmir.IntType(bitwidth))
            desired_arg = builder.bitcast(desired_arg, llvmir.IntType(bitwidth))
```
Then we do not need two separate cases for fp32 and fp64.

numba_dpex/experimental/_kernel_dpcpp_spirv_overloads/_atomic_ref_overloads.py
```diff
@@ -323,6 +324,108 @@ def _intrinsic_exchange_gen(context, builder, sig, args):
...
+        # If the return value is not same as expected, then store return
+        # value in expected.
+        # In either case, return result of cmp instruction.
+        is_cmp_exchg_success = builder.icmp_signed("==", ret_val, expected_arg)
+
+        with builder.if_else(is_cmp_exchg_success) as (then, otherwise):
+            with then:
+                pass
+            with otherwise:
+                if sig.args[0].dtype == types.float32:
```
diptorupd, line 419: Same here.

diptorupd (approved):
Overall looks good. I have one last improvement suggestion and after that the PR is good to go. Thank you @adarshyoga!

numba_dpex/experimental/_kernel_dpcpp_spirv_overloads/_atomic_ref_overloads.py
diptorupd: No, cannot be done here as llvmlite.ir does not expose a single fp type.

numba_dpex/experimental/_kernel_dpcpp_spirv_overloads/_atomic_ref_overloads.py
diptorupd: Let us skip as we have the same pattern in multiple places and let us just be consistent.

PR #1313, opened 2024-02-03
Revert "Remove numba monkey patch that was pushed upstream."

diptorupd:
This reverts commit da1db14220cf5817ca589f94f924aa65a9fa43a2 that removed the monkey patch to numba's mk_alloc function. Without the patch there are failures in the dpjit parfor pipeline with Numba 0.59.

Reverting till we have a better root cause analysis.


diptorupd:
It was some problem with my local env and the PR is not required.

PR #1319, opened 2024-02-06
Capturing return value from fetch_* atomic functions

adarshyoga:

The atomic fetch_* functions return the old value of the atomic reference variable that persisted before the fetch_8 operation is performed. Our implementation was not returning the old value. This PR updates the atomic fetch_* implementations to return the old value.
It also includes updates to the AtomicRef python implementations to return the old value.


diptorupd (approved):
Thank you!

PR #1331, opened 2024-02-13
A sycl::local_accessor-like API for numba-dpex kernel

diptorupd:

The PR adds a `sycl::local_accessor` like API to numba-dpex's kernel API.

TODOs
- [x] All existing test cases pass
- [x] New test case to verify that the kernel function argument for a local_accessor is correctly generated
- [x] dpctl PR merged with new changes [IntelPython/dpctl/pull/1558]
- [x] Remove hard-coding for typeid (after adding a new enum in dpctl) [#1379]
- [x] Add end-to-end tests
- [ ] Example
- [ ] Tutorial
- [x] According to section 4.7.6.11. of the SYCL specification, a local_accessor must not be used in a SYCL kernel function that is invoked via single_task or via the simple form of parallel_for that takes a range parameter.

### DIFF
--- a/numba_dpex/core/types/kernel_api/local_accessor.py
+++ b/numba_dpex/core/types/kernel_api/local_accessor.py
@@ -10,6 +10,15 @@ from numba_dpex.core.types import USMNdArray
 from numba_dpex.utils import address_space as AddressSpace
 
 
+class DpctlMDLocalAccessorType(Type):
+    """numba-dpex internal type to represent a dpctl SyclInterface type
+    `MDLocalAccessorTy`.
+    """
+
+    def __init__(self):
+        super().__init__(name="DpctlMDLocalAccessor")
+
+
 class LocalAccessorType(USMNdArray):
     """numba-dpex internal type to represent a Python object of
     :class:`numba_dpex.experimental.kernel_iface.LocalAccessor`.
--- a/numba_dpex/core/types/kernel_api/local_accessor.py
+++ b/numba_dpex/core/types/kernel_api/local_accessor.py
@@ -2,8 +2,9 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-from numba.core.pythonapi import unbox
-from numba.core.types import Array, Type
+from numba.core import cgutils
+from numba.core.types import Type, UniTuple, intp
+from numba.extending import NativeValue, unbox
 from numba.np import numpy_support
 
 from numba_dpex.core.types import USMNdArray
@@ -69,7 +70,15 @@ def unbox_local_accessor(typ, obj, c):  # pylint: disable=unused-argument
     generate the kernel signature passing in a pointer in the local address
     space.
     """
+    shape = c.pyapi.object_getattr_string(obj, "_shape")
+    local_accessor = cgutils.create_struct_proxy(typ)(c.context, c.builder)
 
-    nparrobj = c.pyapi.object_getattr_string(obj, "_data")
-    nparrtype = Array(typ.dtype, typ.ndim, typ.layout, readonly=False)
-    return c.unbox(nparrtype, nparrobj)
+    ty_unituple = UniTuple(intp, typ.ndim)
+    ll_shape = c.unbox(ty_unituple, shape)
+    local_accessor.shape = ll_shape.value
+
+    return NativeValue(
+        c.builder.load(local_accessor._getpointer()),
+        is_error=ll_shape.is_error,
+        cleanup=ll_shape.cleanup,
+    )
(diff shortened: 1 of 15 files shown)


numba_dpex/kernel_api/local_accessor.py
```diff
@@ -0,0 +1,101 @@
...
+        self._data = numpy.empty(self._shape, dtype=self._dtype)
+
+    def __getitem__(self, idx_obj):
+        """Returns the value stored at the position represented by idx_obj in
+        the self._data ndarray.
+        """
+
+        return self._data[idx_obj]
+
+    def __setitem__(self, idx_obj, val):
```
AlexanderKalistratov, line 86: Why do you need this?
Any attempt to read/write to local memory on host should result in exception.

Or I misunderstands something?

numba_dpex/kernel_api/local_accessor.py
```diff
@@ -0,0 +1,101 @@
...
+            )
+
+        # Make sure shape has a rank between (1..3)
+        if len(self._shape) < 1 or len(self._shape) > 3:
+            raise TypeError("LocalAccessor can only have up to 3 dimensions.")
+
+        self._dtype = dtype
+
+        if self._dtype not in [
+            numpy.float32,
```
AlexanderKalistratov, line 66: is there any reason for such limitation?

numba_dpex/kernel_api/local_accessor.py
diptorupd: Sorry, I missed your comment.

This class is basically a simulator in pure Python for prototyping and testing. Borrowing idea's for numba-mlir's simulator.

The usage is, a `LocalAccessor` is crated in pure Python and passed as a kernel argument. In simulator mode it is just a NumPy array so works as expected. In compiled mode it will be an actual local array. What do you think?

numba_dpex/kernel_api/local_accessor.py
diptorupd: Not really, but what else will be needed?

numba_dpex/kernel_api/local_accessor.py
AlexanderKalistratov: Since there is no examples I'm not sure how it should work, but I assume, that user need to create instance of this class in order to use local memory inside of kernel. Something like that:

```python
loc_mem = LocalAccessor(...)
kernel(..., loc_mem, ...)
```
If that's assumption is correct, then user may try to access local memory outside of kernel:
```python
loc_mem = LocalAccessor(...)
loc_mem[0] = 10
kernel(..., loc_mem, ...)
```

And in this case an exception must be thrown.

numba_dpex/kernel_api/local_accessor.py
AlexanderKalistratov: [u]int8, [u]int16, [u]int32, [u]int64

numba_dpex/experimental/models.py
```diff
@@ -85,3 +88,5 @@ def _init_exp_data_model_manager() -> DataModelManager:
 
 # Register the NdItemType type
 register_model(NdItemType)(EmptyStructModel)
+
+register_model(LocalAccessorType)(ArrayModel)
```
ZzEeKkAa, line 92:
```suggestion
register_model(LocalAccessorType)(EmptyStructModel)
```

numba_dpex/kernel_api/local_accessor.py
ZzEeKkAa: Let's then introduce `LocalAccessorMock` with those methods implemented.

diptorupd: I have made these changes. Can you please review again?

diptorupd:
@ZzEeKkAa please review once again. I have added all changes and tests. I will add an example and update the kernel programming guide in a separate PR.

numba_dpex/core/types/kernel_api/local_accessor.py
```diff
@@ -10,6 +10,15 @@
 from numba_dpex.utils import address_space as AddressSpace
 
 
+class DpCtlMDLocalAccessorType(Type):
+    """numba-dpex internal type to represent a DpCtl SyclInterface type
```
diptorupd, line 15: s/Dpctl/dpctl

numba_dpex/core/utils/kernel_flattened_args_builder.py
```diff
@@ -262,6 +270,75 @@ def _build_local_accessor_metadata_arg(
             ),
         )
 
+    def _build_local_accessor_arg(self, arg_type: LocalAccessorType, llvm_val):
+        """Creates a list of LLVM Values for an unpacked USMNdArray kernel
```
diptorupd, line 274: docstring is wrong as it is not technically a USMNdArray kernel argument

ZzEeKkAa (dismissed):
LGMT! Thank you!

ZzEeKkAa (approved):

PR #1367, opened 2024-02-26
Accept multidimentional atomic indexes

ZzEeKkAa:
Previous implementation used to work only with flat indexes. This one accept multidimentional indexes.

Question: Should we remove default value for the index? It will cause `TypeError` on 2- and 3-dimentional arrays?


diptorupd:
> Question: Should we remove default value for the index? It will cause TypeError on 2- and 3-dimentional arrays?

Yes, I think so. It was a bad design initially also. We should force people to be explicit with creation of atomic ref.

diptorupd (dismissed):
LGTM! Good catch.

diptorupd (approved):

PR #1370, opened 2024-03-01
Add PrivateArray kernel_api

ZzEeKkAa:
Creates `PrivateArray` class that is used inside `kernel_api`. Overloaded with `USMNdArray` instance generation on stack.

### DIFF
--- a/numba_dpex/kernel_api_impl/spirv/arrayobj.py
+++ b/numba_dpex/kernel_api_impl/spirv/arrayobj.py
@@ -1,0 +1,121 @@
+# SPDX-FileCopyrightText: 2024 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Contains SPIR-V specific array functions."""
+
+import operator
+from functools import reduce
+from typing import Union
+
+import llvmlite.ir as llvmir
+from llvmlite.ir.builder import IRBuilder
+from numba.core import cgutils, errors, types
+from numba.core.base import BaseContext
+
+from numba_dpex.kernel_api_impl.spirv.target import SPIRVTargetContext
+from numba_dpex.ocl.oclimpl import _get_target_data
+
+
+def get_itemsize(context: SPIRVTargetContext, array_type: types.Array):
+    """
+    Return the item size for the given array or buffer type.
+    Same as numba.np.arrayobj.get_itemsize, but using spirv data.
+    """
+    targetdata = _get_target_data(context)
+    lldtype = context.get_data_type(array_type.dtype)
+    return lldtype.get_abi_size(targetdata)
+
+
+def require_literal(literal_type: types.Type):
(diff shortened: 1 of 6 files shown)


numba_dpex/kernel_api_impl/spirv/arrayobj.py
```diff
@@ -0,0 +1,117 @@
...
+    ty_array: types.Array,
+    ty_shape: Union[types.IntegerLiteral, types.BaseTuple],
+    shape: llvmir.Value,
+    data: llvmir.Value,
+):
+    """Makes SPIR-V array and fills it data."""
+    # Create array object
+    ary = context.make_array(ty_array)(context, builder)
+
+    itemsize = get_itemsize(context, ty_array)
```
diptorupd, line 55: Did you verify that we do not need the LLVM codegen target for itemsize here?

diptorupd (changes requested):
Few technical questions. The docstrings all have copypasta issues.

numba_dpex/kernel_api_impl/spirv/arrayobj.py
ZzEeKkAa: I'm using the same target as `numba_dpex.private.array`. See the implementation:
```python
def get_itemsize(context: SPIRVTargetContext, array_type: types.Array):
    """
    Return the item size for the given array or buffer type.
    Same as numba.np.arrayobj.get_itemsize, but using spirv data.
    """
    targetdata = _get_target_data(context)
    lldtype = context.get_data_type(array_type.dtype)
    return lldtype.get_abi_size(targetdata)
```

numba_dpex/kernel_api_impl/spirv/arrayobj.py
```diff
@@ -0,0 +1,117 @@
...
+    context: BaseContext,
+    builder: IRBuilder,
+    ty_array: types.Array,
+    ty_shape: Union[types.IntegerLiteral, types.BaseTuple],
+):
+    """Allocates flat array of given shape on the stack."""
+    if not isinstance(ty_shape, types.BaseTuple):
+        ty_shape = (ty_shape,)
+
+    return cgutils.alloca_once(
```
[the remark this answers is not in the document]
ZzEeKkAa: Yes,
```python
def alloca_once(builder, ty, size=None, name='', zfill=False):
    """Allocate stack memory at the entry block of the current function
    pointed by ``builder`` with llvm type ``ty``.  The optional ``size`` arg
    set the number of element to allocate.  The default is 1.  The optional
    ``name`` arg set the symbol name inside the llvm IR for debugging.
    If ``zfill`` is set, fill the memory with zeros at the current
    use-site location.  Note that the memory is always zero-filled after the
    ``alloca`` at init-site (the entry block).
    """
```

diptorupd (approved):
Thank you!

PR #1377, opened 2024-03-06
Fix/private array

ZzEeKkAa:
Use lower instead of overload, since we are allocating memory on the stack. Also adds optional argument `fill_zeros` cause previous implementation was doing so, but it is not essential and overkill for the performance.


diptorupd (dismissed):
Thank you!

numba_dpex/experimental/_kernel_dpcpp_spirv_overloads/_private_array_overloads.py
```diff
@@ -23,55 +24,12 @@
 )
 from numba_dpex.utils import address_space as AddressSpace
 
-from ..target import DPEX_KERNEL_EXP_TARGET_NAME
+from ._registy import lower
```
diptorupd, line 27: Typo in module name.
```suggestion
from ._registry import lower
```

diptorupd (dismissed):
Important set of fixes that need to go in ASAP. Thank you very much!

diptorupd (approved):
good to go!

PR #1379, opened 2024-03-07
Use enum values for kernel arg types if dpctl >= 0.17

diptorupd:
https://github.com/IntelPython/dpctl/pull/1581 updated the enum values for DPCTLSyclKernelArgTy and also introduced a Python enum to track the libsyclinterface values. The current numba-dpex PR uses the latest API and enum from dpctl if the semantic version of dpctl is 0.17 or greater. Old behavior is retained for backward compatibility.


ZzEeKkAa (approved):
LGTM if tests pass!

PR #1382, opened 2024-03-11
Use enum values for kernel arg types if dpctl >= 0.17

ZzEeKkAa:
After https://github.com/IntelPython/dpctl/pull/1585 was merged, we can use enum literals.


diptorupd (approved):
I was about to open the same PR 😄

PR #1391, opened 2024-03-19
Verify global_barrier, indexing, private array inside device_func

diptorupd:

The existing `func` decorator has several design limitation that does not allow usage of features such as `barrier`, `dpex.local.array` and `dpex.private.array` inside a `func` decorated function. These limitations reduced the utility of the feature. See: https://github.com/soda-inria/sklearn-numba-dpex/blob/747914481371ee7fe6527240a7bb57104d207115/sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py#L142

All these limitations are addressed by the new `experimental.device_func` decorator. The PR adds unit tests to verify that `kernel_api.global_barrier` and `kernel_api.PrivateArray` can be used in a `device_func` decorated function.


ZzEeKkAa (approved):
LGTM! Thank you

PR #1394, opened 2024-03-20
Fix KernelHasReturnValueError inside KernelDispatcher.

diptorupd:
The KernelHasReturnValueError check was not working as intended in the experimental module. The PR fixes it by moving the check after compilation and using the signature as derived by the compilation result.


ZzEeKkAa (approved):
LGTM! Nice change!

PR #1397, opened 2024-03-20
Fix use of dpctl.device_context in tests

diptorupd:
We had a left over use of `dpctl.device_context` in our test suite. Now that the `device_context` context manager has been fully removed from dpctl, the public Github CI was failing. The PR updates numba-dpex's source to not use the `device_context`.


ZzEeKkAa (approved):
LGTM!

PR #1405, opened 2024-03-24
Remove/constant size static local memory pass

diptorupd:

The `ConstantSizeStaticLocalmemeoryPass` was an untyped pass in the numba-dpex kernel compiler to verify that any
`dpex.local.array` allocation only used a compile-time constant to specify the size of the array. The restriction was due to the fact that numba-dpex was allocating local memory as a global array in the compiled module.

With `dpex.local.array` not being used and slotted for removal post the `LocalAccessor` feature, the pass is no longer needed.


ZzEeKkAa (approved):
LGTM! Thank you!

PR #1411, opened 2024-03-26
Remove/kernel_interface.spirv_kernel

diptorupd:
The PR does the initial refactoring to move the parfor kernel lowering to new API. The initial changes are as follows:

1. Do not use the `core.kernel_interface.spirv_kernel.SpirvKernel` class in `core.parfors.kernel_builder.py`
2. Move `core.compiler.py` into `core.parfors`. The module is now only used for parfor lowering.
3. Move `core.utils.kernel_templates` into `core.parfors`. It was initially kept as a separate module to possibly use in other places such as vectorize decorator implementation.
4. Delete `core.kernel_interface.kernel_base` and `core.kernel_interface.spirv_kernel` as these modules are no longer used.

No, existing tests suffice.


ZzEeKkAa (approved):
LGMT! Thx

PR #1418, opened 2024-03-30
Docs/fixes to docstrings

diptorupd:
Last few documentation clean up.


numba_dpex/core/runtime/context.py
```diff
@@ -439,16 +439,18 @@ def acquire_meminfo_and_schedule_release(
...
-        DPEXRT_nrt_acquire_meminfo_and_schedule_release(
-            NRT_api_functions *nrt,
-            DPCTLSyclQueueRef QRef,
-            NRT_MemInfo **meminfo_array,
-            size_t meminfo_array_size,
-            DPCTLSyclEventRef *depERefs,
-            size_t nDepERefs,
-            int *status,
-        );
+        .. code-block::
```
oleksandr-pavlyk, line 442:
```suggestion
        .. code-block:: c
```

numba_dpex/core/runtime/context.py
```diff
@@ -475,16 +477,18 @@ def acquire_meminfo_and_schedule_release(
...
-        DPEXRT_build_or_get_kernel(
-            const DPCTLSyclContextRef ctx,
-            const DPCTLSyclDeviceRef dev,
-            size_t il_hash,
-            const char *il,
-            size_t il_length,
-            const char *compile_opts,
-            const char *kernel_name,
-        );
+        .. code-block::
```
oleksandr-pavlyk, line 480:
```suggestion
        .. code-block:: c
```

Default code-block highlighting is for Python

numba_dpex/core/runtime/context.py
```diff
@@ -511,7 +515,9 @@ def build_or_get_kernel(self, builder: llvmir.IRBuilder, args):
     def kernel_cache_size(self, builder: llvmir.IRBuilder):
         """Inserts LLVM IR to call kernel_cache_size.
 
-        size_t DPEXRT_kernel_cache_size();
+        .. code-block::
```
oleksandr-pavlyk, line 518:
```suggestion
        .. code-block:: c
```

docs/source/conf.py
```diff
@@ -27,7 +27,6 @@
 # ones.
 extensions = [
     "sphinx.ext.todo",
-    "sphinx.ext.intersphinx",
```
oleksandr-pavlyk, line 30: Why is it being deleted? It allows to cross reference into `dpnp` project, or C-API of `dpctl` or syclinterface.

diptorupd: It causes a flood of warnings during sphinx compilation. The issue is `autoapi` uses the same template for all files and then intersphinx finds the same ref in multiple files.

numba_dpex/core/runtime/context.py
diptorupd: thanks

docs/source/conf.py
diptorupd: added back

PR #1419, opened 2024-03-30
Refactors and removes core.kernel interface

diptorupd:
The features of `numba_dpex.core.kernel_interface` are now subsumed by the `kernel_api_impl` module and this PR completes the refactoring and removal of the legacy module.
* Move core.kernel_interface.range_overloads to core.overloads
*  Remove kernel_interface.arrayobj
        - Merges core.kernel_interface.arrayobj into kernel_api_impl.spirv.arrayobj
        - Moves arrayobj.populate_array into target.py
        - Updated populate_array to use USMNdArray type.
        - Remove core.kernel_api


numba_dpex/kernel_api_impl/spirv/target.py
```diff
@@ -220,6 +218,75 @@ def _generate_spir_kernel_wrapper(self, func, argtypes):
         module.get_function(func.name).linkage = "internal"
         return wrapper
 
+    def _populate_array(
```
ZzEeKkAa, line 221: I would prefer to be it consistent with numba implementation and implemented as a function in arrayobj.py

ZzEeKkAa (approved):

PR #1426, opened 2024-04-03
Update reduction_template to use kapi.group_barrier

diptorupd:
There is a minor left over issue in the reduction_template that prevents removal of `ocl` module. The PR replaces uses of `dpex.barrier` with `kernel_api.group_barrier`.


ZzEeKkAa (approved):
Nice catch, thank you!

PR #1427, opened 2024-04-03
Fix test case to work with NUMBA_CAPTURED_ERRORS=new_style

diptorupd:
Update unit test so that it works in `new_style` Numba error capture mode.


ZzEeKkAa (approved):
LGTM!

PR #1430, opened 2024-04-05
fix windows wheels copy cmd

ekomarova:
This PR fixes the windows wheel copying by returning it to the previous cmd. We don't need to store anything in the artifacts except the wheel package. Also, for some reason, cmd does not work correctly and does not save anything in the artifacts at all and the wheel package is missing


ekomarova:
Now windows wheel is present in the artifacts, as it should be

ZzEeKkAa (approved):
Thank you! It looks like copy supports wildcards in different way from `.bat` default behavior.  Good catch

PR #1431, opened 2024-04-05
Update example to use latest dpctl API.

diptorupd:
The `matmul` examples fails in CI as dpctl was updated to a new API. The PR fixes the example by using latest dpctl API.


oleksandr-pavlyk (approved):
Thanks @diptorupd !

PR #1435, opened 2024-04-11
Migrate parfor to SPIRVKernelDispatcher

ZzEeKkAa:
What was done:
- Compile parfor kernels using `SpirvDispatcher`.
- Add internal option to compile kernel with Prange support. This mode inserts prange body instead of `__seninel__ = 0`. Ideally it  must be replace to some internal dedicated function call like `numba_dpex.core.insert_parfor_body()`.
- Remove kernel allocation aside of dpjit compilation. Now parfor kernel specific SPIRV is a part of dpjit function inserted as a string constant. It unblocks dpjit function caching on disk

Checklist:

### DIFF
--- a/numba_dpex/core/parfors/kernel_builder.py
+++ b/numba_dpex/core/parfors/kernel_builder.py
@@ -26,8 +26,17 @@ from numba.core.typing import signature
 from numba.parfors import parfor
 
 from numba_dpex.core import config
+from numba_dpex.core.decorators import kernel
+from numba_dpex.core.parfors.parfor_sentinel_replace_pass import (
+    ParforBodyArguments,
+)
 from numba_dpex.core.types.kernel_api.index_space_ids import ItemType
+from numba_dpex.core.utils.call_kernel_builder import SPIRVKernelModule
 from numba_dpex.kernel_api_impl.spirv import spirv_generator
+from numba_dpex.kernel_api_impl.spirv.dispatcher import (
+    SPIRVKernelDispatcher,
+    _SPIRVKernelCompileResult,
+)
 
 from ..descriptor import dpex_kernel_target
 from ..types import DpnpNdArray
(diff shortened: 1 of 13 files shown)


numba_dpex/core/parfors/kernel_builder.py
```diff
@@ -367,120 +249,33 @@ def create_kernel_for_parfor(
...
-            print("No aliases found so adding noalias flag.")
-        flags.noalias = True
-
-    remove_dead(kernel_ir.blocks, kernel_ir.arg_names, kernel_ir, typemap)
-
-    if config.DEBUG_ARRAY_OPT:
-        print("kernel_ir after remove dead")
-        kernel_ir.dump()
-
-    # The first argument to a range kernel is a kernel_api.Item object. The
```
diptorupd, line 429: Maybe keep the code comment around.

numba_dpex/core/parfors/kernel_parfor_pass.py
```diff
@@ -0,0 +1,103 @@
+from typing import NamedTuple, Union
```
diptorupd, line 1: license header is missing

diptorupd: I also think `parfor_kernel_pass` or `parfor_to_kernel_pass` is a better name as a parfor is being converted to a kernel.

ZzEeKkAa: I guess `parfor_inject_kernel_pass` is a better name

numba_dpex/core/pipelines/kernel_compiler.py
```diff
@@ -56,6 +58,16 @@ def define_nopython_pipeline(state, name="dpex_kernel_nopython"):
         ndpb = DefaultPassBuilder
         pm = PassManager(name)
         untyped_passes = ndpb.define_untyped_pipeline(state)
+
+        for i, _pass in enumerate(untyped_passes.passes):
```
ZzEeKkAa, line 62: TODO:
pm.add_pass_after(MlirBackend, NopythonTypeInference)

ZzEeKkAa: Move it into separate pass.

diptorupd (approved):
There are few cleanups such as introducing a new pipeline fr parfor kernel compilation that should be done, can be handled in later PR.

Overall nice set of changes. The parfor kernel generator has become much cleaner after these changes.

PR #1439, opened 2024-04-17
Revert 98f3b05df to see if sources are visible on coveralls.io

diptorupd:
Reverts 98f3b05df  Coveralls.io does not display the sources for numba-dpex anymore. The PR triages if the problem is due to the official coveralls workflow.

Although not showing sources is a feature and not a bug (https://github.com/lemurheavy/coveralls-public/issues/1579), the sources are displayed for dpctl and dpnp. So, let us see what happens if we go back to what was there before.


diptorupd:
@ZzEeKkAa Want your thoughts on this PR. The issue is after switching to official action, coveralls.io does not show any source files. Reverting the change does show the sources.

PR #1446, opened 2024-04-30
Fix coveralls sources

ZzEeKkAa:
Add source directory, so the generated xml file has the same root directory as a package repo.

Closes #1439


diptorupd (approved):
Thank you!

PR #1447, opened 2024-05-01
Remove artifacts from conda/wheel packages residing in root level

AndresGuzman-Ballen:
Removing files that are making their way to conda packages at root-level.

PR #1450, opened 2024-05-06
Remove MANIFEST.in

ZzEeKkAa:
Reverts #1447 and fixes root cause of the problem: Manifest was setup incorrectly, that resulted wheel to install some artifacts in the root directory. It is no longer required as it managed through the setuptools configuration directly.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,0 @@
-include MANIFEST.in
-include README.md setup.py LICENSE
-
-include numba_dpex/_version.py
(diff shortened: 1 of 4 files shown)


AndresGuzman-Ballen:
Whoops! My bad :(

Don't forget to remove the changes from bld.bat too :)

ZzEeKkAa:
Looks like it is not that simple..

diptorupd (approved):

PR #1452, opened 2024-05-07
update wheel tag with pep600

ekomarova:
This PR is designed to fix the problem described here: [SAT-6873](https://jira.devtools.intel.com/browse/SAT-6873). I did some experiments and found out that using pep600 wheel tag prevents the wheel from being installed on non-supported platforms


ekomarova:
3.11 failed as expected because dpcpp-llvm-spirv 2024.0 is not built for 3.11 and this causes conflicts (will be fixed later)

ZzEeKkAa (approved):
LGTM! I've updated the PR to use glibc version from system, not like a magic constant

diptorupd (approved):

PR #1454, opened 2024-05-09
Fixed ref-counting of Python object temporaries in unboxing code

oleksandr-pavlyk:
Ref-count of `arrayobj` was not decremented after use in case `dpnp.ndarray` object was being processed.

Change the flow of managing lifetime of Python temporaries in unboxing code.
This PR replaces use of `obj` in NRT structure with use of `arrayobj`, and shifts ref-count manipulation of `obj` into `PyUSMNdArray_ARRAYOBJ`.

### DIFF
--- a/numba_dpex/core/runtime/_dpexrt_python.c
+++ b/numba_dpex/core/runtime/_dpexrt_python.c
@@ -744,14 +744,18 @@ static struct PyUSMArrayObject *PyUSMNdArray_ARRAYOBJ(PyObject *obj)
         DPEXRT_DEBUG(
             drt_debug_print("DPEXRT-DEBUG: usm array was passed directly\n"));
         arrayobj = obj;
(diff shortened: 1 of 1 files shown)


diptorupd:
@oleksandr-pavlyk there are some compile time warnings after these changes.

adarshyoga:
The memory leak that was occurring with pairwise distance workload is resolved with this fix. Thanks, @oleksandr-pavlyk .

oleksandr-pavlyk:
> @oleksandr-pavlyk there are some compile time warnings after these changes.

Yes, I pushed a fix for them, by casting `PyUSMArrayObject *` pointer to `PyObject *` pointer, which is always safe to do.

ZzEeKkAa (approved):

diptorupd (approved):

PR #1458, opened 2024-05-14
Use nrt api to allocate meminfo object

ZzEeKkAa:
Replace `NRT_MemInfo` allocation from std's `malloc` to NRT API, so they handle the same way for the whole lifetime. Was causing crashes under expect tests

### DIFF
--- a/numba_dpex/core/runtime/_dpexrt_python.c
+++ b/numba_dpex/core/runtime/_dpexrt_python.c
@@ -45,12 +45,14 @@ static NRT_MemInfo *DPEXRT_MemInfo_fill(NRT_MemInfo *mi,
                                         bool value_is_float,
                                         int64_t value,
                                         const DPCTLSyclQueueRef qref);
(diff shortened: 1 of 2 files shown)


numba_dpex/core/runtime/_dpexrt_python.c
```diff
@@ -428,7 +437,8 @@ static NRT_MemInfo *NRT_MemInfo_new_from_usmndarray(PyObject *ndarrobj,
     DPCTLSyclContextRef cref = NULL;
 
     // Allocate a new NRT_MemInfo object
-    if (!(mi = (NRT_MemInfo *)malloc(sizeof(NRT_MemInfo)))) {
+    // By passing 0 we are just allocating MemInfo
```
diptorupd, line 440:
```suggestion
    // By passing 0 we are just allocating MemInfo and not the `data` that the MemInfo object manages.
    // The `data` object is allocated separately, if needed, using a USM allocator.
```

diptorupd (approved):
Thank you! Looks great barring minor nit.

diptorupd (approved):

PR #1464, opened 2024-05-23
Remove versioneer, organize dpnp monkey patch

ZzEeKkAa:
Moves out dpnp user functions monkey patching to the designed patch folder. Updates `_versioner.py` on the build time so it is statically set and does not trigger any subprocesses, like git.


numba_dpex/core/typing/dpnpdecl.py
```diff
@@ -25,7 +22,7 @@ def _install_operations(cls: NumpyRulesArrayOperator):
     for op, ufunc_name in cls._op_map.items():
         infer_global(op)(
             type(
-                "NumpyRulesArrayOperator_" + ufunc_name,
+                "DpnpRulesArrayOperator_" + ufunc_name,
```
diptorupd, line 25: :smile_cat:

diptorupd (approved):
Thank you for the changes!

